PR #30, opened 2020-03-12
update COBRAsolver to add weighted constraints

hemilpanchiwala:
In COBRAsolver class, when any reactant is just involved in a single reaction, then there should be no weight constraint for it. 

This pull request solves this issue by adding the condition to add the weight constraint when the reactant is involved in more than one reaction.

### DIFF
--- a/src/main/java/org/simulator/fba/COBRAsolver.java
+++ b/src/main/java/org/simulator/fba/COBRAsolver.java
@@ -238,6 +238,7 @@ public class COBRAsolver {
 		// Add weighted constraints equations for each reaction.
 		for (Species species : model.getListOfSpecies()) {
 			double[] weights = new double[reaction2Index.size()];
+			int numberOfReactions = 0;
 			
 			if (!species2Reaction.containsKey(species.getId())) {
 				logger.warning(MessageFormat.format(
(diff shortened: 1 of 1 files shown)


src/main/java/org/simulator/fba/COBRAsolver.java
```diff
@@ -246,10 +247,13 @@ public COBRAsolver(SBMLDocument doc) throws SBMLException, ModelOverdeterminedEx
 			} else {
 				for (Pair<String, Double> pair : species2Reaction.get(species.getId())) {
 					weights[reaction2Index.get(pair.getKey())] = pair.getValue();
+					numberOfReactions++;
```
draeger: The number of reactions in which a species participates can be obtained without the need to add another counter. You can just get it from `species2Reaction.get(species.getId()).size()`.

src/main/java/org/simulator/fba/COBRAsolver.java
```diff
@@ -246,10 +247,13 @@ public COBRAsolver(SBMLDocument doc) throws SBMLException, ModelOverdeterminedEx
...
 				}
 			}
 			
-			problem.addConstraint(new LinearEqualsConstraint(weights, 0.0, "cnstrt_" + species.getId()));
```
draeger: Could you please elaborate a bit more what the problem is when a species participates in only one reaction?

hemilpanchiwala: @draeger, when the species participates in only one reaction (let it be R1), then the constraint R1 = 0 will be added to the linear problem which does not make sense and also, leads to the wrong solution.

I found this error while running the FBA for the example given in the FBC specification document.

hemilpanchiwala:
> Thanks for doing this fix! According to Travis there is now a build error. Do you know where this is coming from?

@draeger, this error is occurring due to some bug in the CompSimulator. The test files for comp are not executing successfully. 

The error message in the image occurs while running the simulation for `src/test/resources/comp/test6.xml`


Similar error is for `test7.xml` to `test10.xml` files.

src/main/java/org/simulator/fba/COBRAsolver.java
hemilpanchiwala: @draeger, Updated in the next commit

PR #40, opened 2020-06-03
Resolves issues of SBMLTestSuite and FBA

hemilpanchiwala:
Major changes from this PR:
- Adds a running wrapper for the SBML Test Runner that allows to check the results through the plot and compare the distance plots.
- Adds testing by CompSimulator for the comp models in SBMLTestSuiteTest
- Adds three new distance metrics in SBSCL.
	- Maximum Absolute Distance
	- Maximum Relative Distance
	- Maximum Divergence Tolerance - The metric used by SBML Test Runner
- Adds the newly added tests from SBML Test Suite.
- Updates the INSTALL.md section for the SBML Test Runner.
- Moves all the script files under the `src/test/scripts` folder. Updated the scripts files.
- CobraSolver is refactored to FluxBalancedAnalysis.
- Solves the issues with the FluxBalanceAnalysis so that all the FBC tests from SBML Test Suite now runs perfectly.
- Fixes some other minor issues and some enhancement in the code


src/main/java/org/simulator/math/QualityMeasure.java
```diff
@@ -160,14 +161,21 @@ public double distance(MultiTable x, MultiTable expected) {
...
     return meanFunction.computeMean(distances);
   }
 
+  /**
+   *
+   * @param x
+   * @param expected
+   * @return maxAbsDistances the list of maximum absolute distances in the blocks
+   */
   public ArrayList<Double> getMaxAbsDistances(MultiTable x, MultiTable expected) {
```
draeger: Please do not declare unnecessarily specific return types for methods and also declare variables with the least specific type. In this concrete case, please use `List<Double>` to initialize the variable for the `distances` and also return a `List<Double>`. You can still use an `ArrayList<Double>` to initialize the variable, but not as return type. This applies to other positions as well.

src/test/download_bigg_models.sh
```diff
@@ -6,11 +6,13 @@
...
+_CWD="$PWD"
 TEST_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 # download and extract the bigg models for testing
-mkdir -p resources/bigg
-cd resources/bigg
+cd TEST_DIR
+mkdir -p $TEST_DIR/resources/bigg
+cd $TEST_DIR/resources/bigg
 wget https://github.com/matthiaskoenig/bigg-models-fba/raw/master/models/bigg_models_v1.5.tar.gz
```
draeger: @matthiaskoenig Not clear to me why we don't download the BiGG models directly from BiGG. Doesn't this detour introduce the problem that we need to keep two places in sync with BiGG Models Database?

src/test/java/org/simulator/sbml/SBMLTestSuiteTest.java
```diff
@@ -144,50 +147,77 @@ public void testModel() throws FileNotFoundException, IOException {
...
+                            errorInSolve = true;
+                        }
+                        Assert.assertNotNull(solution);
+                        Assert.assertFalse(errorInSolve);
+                        Assert.assertFalse(solver.isUnstable());
+
+                        // compute distance
+                        QualityMeasure distance = new RelativeEuclideanDistance();
+                        double dist = distance.distance(solution, inputData);
+                        Assert.assertTrue(dist <= 0.2);
```
draeger: I'd recommend to always indicate the data type also for literal numbers, here, for instance, `0.2d` so that the compiler knows it is a `double` value. If you do it always, you will not forget it in situations where it can make a difference (e.g., when dividing integer numbers).

src/main/java/org/simulator/math/QualityMeasure.java
```diff
@@ -159,73 +160,57 @@ public double distance(MultiTable x, MultiTable expected) {
...
-   * 
+   *
    * @param x
    * @param expected
-   * @return columnDistances the list of distances for the columns in the blocks
+   * @return maxAbsDistances the list of maximum absolute distances in the blocks
    */
-  public ArrayList<Double> getColumnDistances(MultiTable x, MultiTable expected) {
-    ArrayList<Double> distances= new ArrayList<Double>();
+  public HashMap<String, Double> getMaxAbsDistances(MultiTable x, MultiTable expected) {
```
draeger, line 199: Also here, please do not return `HashMap<String, Double>` but `Map<String, Double>` to make it easier to use a different implementation later on.

src/test/java/org/simulator/sbml/SBMLTestSuiteTest.java
hemilpanchiwala: @draeger, this has been removed from the code now. The distance has been now updated to the MaxDivergenceTolerance.

hemilpanchiwala:
> As a general recommendation, I'd suggest making use of the templates for code style from the JSBML project at https://github.com/sbmlteam/jsbml/tree/master/dev/. They can really help to automatically insert blank spaces, curly braces, and so forth, where forgotten.
> 
> Further comments were given in the code.

I can just see their recommendations but can't add the templates as I am working on IntelliJ (even prefer), and the templates are only present for Eclipse.

draeger:
> > As a general recommendation, I'd suggest making use of the templates for code style from the JSBML project at https://github.com/sbmlteam/jsbml/tree/master/dev/. They can really help to automatically insert blank spaces, curly braces, and so forth, where forgotten.
> > Further comments were given in the code.
> 
> @draeger, I can just see their recommendations but can't add the templates as I am working on IntelliJ (even prefer), and the templates are only present for Eclipse.

Ok; so IntelliJ doesn't have an import for such files or doesn't come with comparable functions?

hemilpanchiwala:
> > > As a general recommendation, I'd suggest making use of the templates for code style from the JSBML project at https://github.com/sbmlteam/jsbml/tree/master/dev/. They can really help to automatically insert blank spaces, curly braces, and so forth, where forgotten.
> > > Further comments were given in the code.
> > 
> > 
> > @draeger, I can just see their recommendations but can't add the templates as I am working on IntelliJ (even prefer), and the templates are only present for Eclipse.
> 
> Ok; so IntelliJ doesn't have an import for such files or doesn't come with comparable functions?

I can just format the code as per the Java code style which is provided in IntelliJ given at this [link](https://www.jetbrains.com/help/idea/reformat-and-rearrange-code.html).

src/main/java/org/simulator/math/QualityMeasure.java
hemilpanchiwala: @draeger, Updated this in the latest commit.

.travis.yml
```diff
@@ -5,12 +5,12 @@ jdk:
   - oraclejdk11
 
 install:
-   - cd src/test
+   - cd src/test/scripts
```
matthiaskoenig, line 8: Don't do the cd and back here!
Just directly call the scripts as mentioned in the installation instructions
```
source ./src/test/scripts/download_bigg_models.sh
source ./src/test/scripts/download_sbml-test-suite.sh
```

INSTALL.md
```diff
@@ -1,4 +1,38 @@
...
+
+SBSCL can be build with maven. To build without running the tests use
+```
+mvn clean install -DskipTests
+```
+
+The tests require additional resources which have to be downloaded. These include the sbml-testsuite and the BiGG models.
+```
+source ./src/test/download_bigg_models.sh
+source ./src/test/download_sbml-test-suite.sh
```
matthiaskoenig, line 13: missing the `scripts` directory

INSTALL.md
```diff
@@ -1,4 +1,38 @@
...
+```
+which then can be run as part of the build step
+```
+mvn clean install
+```
+
+## SBML Test Runner
+The SBML Test Runner is a standalone desktop application with a graphical user interface (GUI) for controlling SBML-compatible applications and making them run test cases from the [SBML Test Suite](http://sbml.org/Software/SBML_Test_Suite). It is written in Java with the Eclipse SWT widget toolkit and can be used on macOS, Linux and Windows.
```
matthiaskoenig, line 21: Delete: `It is written in Java with the Eclipse SWT widget toolkit and can be used on macOS, Linux and Windows.`

INSTALL.md
```diff
@@ -1,4 +1,38 @@
...
+- Then just add the Wrapper path (absolute path to `sbml_test_suite_runner_wrapper.sh` present in SBSCL source code) and output directory path (wherever you want to store the results of the simulator).
+- And, just click on Save and it's done!
+- Now, you can simulate any of the SBML Test Suite case and see the results and the distance plot (in reference to predefined solutions from SBML Test Suite) in the test runner.
+- The distance plots in the test runner is calculated under the criteria given at this [link](http://sbml.org/Software/SBML_Test_Suite/Case_Descriptions#The_.22settings.22_file).
+
+For complete guidelines about the SBML Test Runner and the wrapper, go to the [SBML Test Runner repository](https://github.com/sbmlteam/sbml-test-suite/tree/master/src/test-runner/testsuite-ui).  
+
 
 ## LP solver
 For running flux balance analysis simulations a LP solver is required. The respective jars have to be made available.
```
matthiaskoenig, line 38: Can you add a bit more information for the LP solvers. How did you install them? Where did you download what?
> ## LP solver
> For running flux balance analysis simulations a LP solver is required. The respective jars have to be made available.

docs/apidocs/org/simulator/examples/COBRAsolverExample.html
```diff
@@ -125,7 +125,7 @@ <h2 title="Class COBRAsolverExample" class="title">Class COBRAsolverExample</h2>
 <li><a href="https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html?is-external=true" title="class or interface in java.lang">java.lang.Object</a></li>
```
matthiaskoenig, line 125: Filename wrong? The files are still called COBRAsolverExample (see also next files below)

src/main/java/org/simulator/examples/FBAExample.java
```diff
@@ -52,12 +52,12 @@
  * @author Shalin Shah
```
matthiaskoenig, line 52: Remove author information. Are we having author information on files in general? I find this unnecessary. All changes are tracked in git. So very clear who did what. 
Please ignore this if author information exists everywhere.

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
```diff
@@ -0,0 +1,180 @@
...
+    public static void main(String[] args) throws IOException, XMLStreamException, ModelOverdeterminedException, DerivativeException {
+
+        // Configuration
+        String dirPath = args[0];
+        String currentCase = args[1];
+        String outputDirPath = args[2];
+        String level = args[3];
+        String  version = args[4];
+
+        String filePath = dirPath + '/' + currentCase + '/' + currentCase + "-sbml-l" + level + 'v' + version + ".xml";
```
matthiaskoenig, line 53: I am pretty sure this is not how file paths should be generated in Java! This will most likely break on some other os then Linux. I think one should use `File.seperator` (at least something like this is done in python and other languages). Please check how to create correct paths in Java.

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
```diff
@@ -0,0 +1,180 @@
...
+                amountHash.put(s, true);
+            }
+        }
+
+        for (String s : concentrations) {
+            s = s.trim();
+            if (!s.isEmpty()) {
+                amountHash.put(s, false);
+            }
+        }
```
matthiaskoenig, line 79: You putting the concentrations in the amountHash? Is this correct?

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
```diff
@@ -0,0 +1,180 @@
...
+        // get timepoints
+        CSVImporter csvimporter = new CSVImporter();
+        MultiTable inputData = csvimporter.convert(model, resultsPath);
+        double[] timepoints = inputData.getTimePoints();
+        duration = timepoints[timepoints.length - 1]
+                - timepoints[0];
+
+        MultiTable solution;
+
+        if (model.getExtension(CompConstants.shortLabel) == null){
```
matthiaskoenig, line 95: You do not set the tolerances anywhere for the simulation. Please set the tolerances explicitly! Only then it is possible to check the solutions correctly.

src/main/java/org/simulator/fba/FluxBalanceAnalysis.java
```diff
@@ -418,4 +450,22 @@ private double stoichiometry(SpeciesReference specRef) throws SBMLException, Mod
 		return Double.NaN;
 	}
 
+	/**
+	 * Gets the value of the EPSILON
+	 *
+	 * @return the epsilon value
+	 */
+	public double getEpsilon() {
+		return EPS;
```
matthiaskoenig, line 459: Please make this lowercase and call it `epsilon` or `eps`.

src/main/java/org/simulator/io/CSVImporter.java
```diff
@@ -104,9 +104,9 @@ public MultiTable convert(Model model, String pathname) throws IOException {
 
       Map<String, Integer> nameToColumn = new HashMap<String, Integer>();
       i = 0;
-      for (String head : header) {
-        if (!head.equalsIgnoreCase(data.getTimeName())) {
-          newHead[i++] = head.trim();
+      for (int p = 1; p < header.length; p++) {
```
matthiaskoenig, line 107: Please add a comment why this loop is starting at 1. What are you ignoring?

src/main/java/org/simulator/math/MaxDivergenceTolerance.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    public MaxDivergenceTolerance(double absTol, double relTol) {
+        super(Double.NaN);
+        metric = new MaxAbsDistance();
+        this.absTol = absTol;
+        this.relTol = relTol;
+    }
+
+    /* (non-Javadoc)
+     * @see org.sbml.simulator.math.Distance#distance(java.lang.Iterable, java.lang.Iterable, double)
```
matthiaskoenig, line 33: Document this better ! Please describe what the distance is and state the used formula in the comment. This is the core comparison function metric used. This must be documented so that others understand what is done here.

src/main/java/org/simulator/math/MaxDivergenceTolerance.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    @Override
+    public double distance(Column x, Column y, double defaultValue) {
```
matthiaskoenig, line 36: Also did not see any test code for this.

src/main/java/org/simulator/fba/FluxBalanceAnalysis.java
draeger: @matthiaskoenig is here referring to the name of the variable, not the name of the method.

hemilpanchiwala:
> change name of method. This is very unspecific. Use something like `addBoundsEpsilon` which states what the method is doing. `updateBounds` could do anything with the bounds. Also probably a good idea to define the eps locally here (because it is not used anywhere else).

@matthiaskoenig, but we have given get and set methods for the eps. So, I don't think we should declare it locally in the method.

src/main/java/org/simulator/io/CSVImporter.java
hemilpanchiwala: @matthiaskoenig, the time column has to be ignored for the newHead array. So, its started at index 1.

hemilpanchiwala: This was the main reason why the travis ignored the SBMLTestSuiteTest file.

src/main/java/org/simulator/math/MaxDivergenceTolerance.java
hemilpanchiwala: Adding in the next commit!

src/main/java/org/simulator/io/CSVImporter.java
draeger: Please just add a comment. Here, it can be as simple as adding `// we are starting at 1 because...` in the line over the loop.

.travis.yml
```diff
@@ -5,12 +5,12 @@ jdk:
...
    # get BiGG models for testing
    - source ./download_bigg_models.sh
    # get sbml-test-case models for testing
    - source ./download_sbml-test-suite.sh
-   - cd ../..
+   - cd ../../..
 
 script:
 - mvn clean install
```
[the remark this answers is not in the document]
hemilpanchiwala: Done!

INSTALL.md
```diff
@@ -1,4 +1,38 @@
 # Installation
+This document provides important information on how to setup and install SBSCL and its optional dependencies.
+
+
+SBSCL can be build with maven. To build without running the tests use
+```
+mvn clean install -DskipTests
+```
+
+The tests require additional resources which have to be downloaded. These include the sbml-testsuite and the BiGG models.
```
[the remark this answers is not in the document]
hemilpanchiwala: Done!

.travis.yml
hemilpanchiwala: Done!

INSTALL.md
hemilpanchiwala: Updated!

INSTALL.md
hemilpanchiwala: Updated!

INSTALL.md
hemilpanchiwala: The link to SCPSolver provides all the information for this. However, I have added information here also.

src/main/java/org/simulator/examples/FBAExample.java
hemilpanchiwala: I found it in most of the classes, so am not removing it.

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
hemilpanchiwala: Updated using `File.separator`!

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
hemilpanchiwala: Updated!

src/main/java/org/simulator/fba/FluxBalanceAnalysis.java
hemilpanchiwala: Updated!

src/main/java/org/simulator/io/CSVImporter.java
hemilpanchiwala: @draeger, this is commented above at line 100 in this file: `// exclude time column`.

src/main/java/org/simulator/math/MaxDivergenceTolerance.java
hemilpanchiwala: Added the comments properly in the initial description of the class.

src/main/java/org/simulator/math/MaxDivergenceTolerance.java
hemilpanchiwala: Added the tests for `MaxDivergenceTolerance` class

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
hemilpanchiwala: amountHash keeps track of the following:
- Is true for those variables whose output in results file is in amount units
- Is false for those variables whose output in results file is in concentration units

It's a String to Boolean hashmap.

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
```diff
@@ -50,10 +48,10 @@ public static void main(String[] args) throws IOException, XMLStreamException, M
...
         String  version = args[4];
 
-        String filePath = dirPath + '/' + currentCase + '/' + currentCase + "-sbml-l" + level + 'v' + version + ".xml";
-        String settingsPath = dirPath + '/' + currentCase + '/' + currentCase + "-settings.txt";
-        String outputFilePath = outputDirPath + '/' + currentCase + ".csv";
-        String resultsPath = dirPath + '/' + currentCase + '/' + currentCase + "-results.csv";
+        String filePath = dirPath + File.separator + currentCase + File.separator + currentCase + "-sbml-l" + level + 'v' + version + ".xml";
+        String settingsPath = dirPath + File.separator + currentCase + File.separator + currentCase + "-settings.txt";
+        String outputFilePath = outputDirPath + File.separator + currentCase + ".csv";
+        String resultsPath = dirPath + File.separator + currentCase + File.separator + currentCase + "-results.csv";
```
draeger, line 54: Did anybody test this under Windows? To my knowledge, internally Java always uses `/` as separator in paths and the virtual machine translates it to the operating system.

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
```diff
@@ -63,6 +61,8 @@ public static void main(String[] args) throws IOException, XMLStreamException, M
         String[] amounts = String.valueOf(properties.getProperty("amount")).split(",");
         String[] concentrations = String.valueOf(
                 properties.getProperty("concentration")).split(",");
+        double absolute = (!properties.getProperty("absolute").isEmpty()) ? Double.parseDouble(properties.getProperty("absolute")) : 0d;
+        double relative = (!properties.getProperty("relative").isEmpty()) ? Double.parseDouble(properties.getProperty("relative")) : 0d;
```
draeger, line 65: It would be preferable to declare `String` literals in the form `public static final String ABSOLUTE = "absolute"` and then to reuse these constants where needed. This reduces maintenance effort and possible errors caused by typos. 

Another possibility is creating an internal `enum` such as:
```
public enum SettingsKeywords {
   absolute;
   ...
}
```
that can be uses similarly.

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
draeger: Just to add: I think this will work, but I also think it is not needed (without having it tested).

hemilpanchiwala: Yes, I just saw that when we give `/`, then Java internally handles things to translate to platform-specific code. But `File.separator` can be a good way to show users what will make sense in their OS, rather than what makes sense to Java.

src/main/java/org/simulator/examples/SBMLTestSuiteRunnerWrapper.java
hemilpanchiwala: Updated by adding final variables for Strings.

PR #54, opened 2020-07-28
Support for stochastic simulation added

hemilpanchiwala:
Updates from this PR:
- The new test file `StochasticTestSuiteTest` is added to SBSCL which simulates all the stochastic tests from the SBML Test Suite.
- Complete output format of FERN is now updated according to the test suite result format and all the tests now simulate fine.
- 36 out of 39 tests now pass successfully (others fail as currently there is no support for rules and species with no substance units).
- Also, as stochastic tests can fail sometimes due to randomness, so for passing the Travis CI, seed values for all the models have been collected and currently, tests are ran using these seed values, so tests run fast and pass successfully.
- Minor fixes in CSVImporter and issues in test cases with the event (there was a triggering issue) that has been resolved now.
- Repetitive initialization in the SBMLinterpreter class is removed.

With these updates and PR #53, stochastic simulation support is now added to SBSCL. 
All the updates of the project are present at my blog posts at this [link](https://gsoc20hemil.blogspot.com/).


zakharc:
@hemilpanchiwala: Please clarify:
> 36 out of 39 tests now pass successfully (others fail as currently there is no support for rules and species with no substance units).
> Also, as stochastic tests can fail sometimes due to randomness, so for passing the Travis CI, seed values for all the models have been collected, and currently, tests are ran using these seed values, so tests run fast and pass successfully.

I see 5 **errors** (not fails) with NPE in the build log:
```
[[1;31mERROR[m] testBasicSEDMLExecutorForMiriamURNDefinedModel(org.simulator.sedml.SEDMLExecutorTest)  Time elapsed: 2.067 s  <<< ERROR!
java.lang.NullPointerException
	at org.simulator.sedml.SEDMLExecutorTest.testBasicSEDMLExecutorForMiriamURNDefinedModel(SEDMLExecutorTest.java:137)

[[1;31mERROR[m] testLeloupSBML(org.simulator.sedml.SEDMLExecutorTest)  Time elapsed: 1.024 s  <<< ERROR!
java.lang.NullPointerException
	at org.simulator.sedml.SEDMLExecutorTest.testSpecificationExample(SEDMLExecutorTest.java:228)
	at org.simulator.sedml.SEDMLExecutorTest.testLeloupSBML(SEDMLExecutorTest.java:153)

[[1;31mERROR[m] testRepressilator1(org.simulator.sedml.SEDMLExecutorTest)  Time elapsed: 0.138 s  <<< ERROR!
java.lang.NullPointerException
	at org.simulator.sedml.SEDMLExecutorTest.testSpecificationExample(SEDMLExecutorTest.java:228)
	at org.simulator.sedml.SEDMLExecutorTest.testRepressilator1(SEDMLExecutorTest.java:249)

[[1;31mERROR[m] testRepressilator(org.simulator.sedml.SEDMLExecutorTest)  Time elapsed: 0.057 s  <<< ERROR!
java.lang.NullPointerException
	at org.simulator.sedml.SEDMLExecutorTest.testSpecificationExample(SEDMLExecutorTest.java:228)
	at org.simulator.sedml.SEDMLExecutorTest.testRepressilator(SEDMLExecutorTest.java:200)

[[1;31mERROR[m] testIkappab(org.simulator.sedml.SEDMLExecutorTest)  Time elapsed: 0.03 s  <<< ERROR!
java.lang.NullPointerException
	at org.simulator.sedml.SEDMLExecutorTest.testSpecificationExample(SEDMLExecutorTest.java:228)
	at org.simulator.sedml.SEDMLExecutorTest.testIkappab(SEDMLExecutorTest.java:147)
```
Also, tons of warnings of this type:
```
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.5: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.6: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
WARNING: [VIOLATION]	(k1 < 10) && (k1 > 0) at time 6.4: 
Jul 30, 2020 3:28:58 AM org.simulator.sbml.SimpleConstraintListener processViolation
```
Or this:
```
 WARN (SBMLReader.java:1096) - Cannot find a parser for the  namespace
 WARN (SBMLReader.java:1324) - !!! event.isEndElement: there is a problem in your SBML file !!!!
 WARN (SBMLReader.java:1325) - This should never happen, there is probably a problem with the parsers used.
 Try to check if one needed parser is missing or if you are using a parser in development.
 WARN (SBMLReader.java:1096) - Cannot find a parser for the  namespace
 WARN (SBMLReader.java:1096) - Cannot find a parser for the  namespace
 WARN (SBMLReader.java:808) - Some characters cannot be read: type
 WARN (SBMLReader.java:1324) - !!! event.isEndElement: there is a problem in your SBML file !!!!
 WARN (SBMLReader.java:1325) - This should never happen, there is probably a problem with the parsers used.
 Try to check if one needed parser is missing or if you are using a parser in development.
 WARN (SBMLReader.java:808) - Some characters cannot be read:  Status report
 WARN (SBMLReader.java:1324) - !!! event.isEndElement: there is a problem in your SBML file !!!!
 WARN (SBMLReader.java:1325) - This should never happen, there is probably a problem with the parsers used.
 Try to check if one needed parser is missing or if you are using a parser in development.
```
Is this expected?

[log.txt](https://github.com/draeger-lab/SBSCL/files/5010255/log.txt)

hemilpanchiwala:
@zakharc, the reason for Travis CI failing currently is that for the simulation of the SED-ML models, it downloads the models via a URN from biomodels. But currently, the site from which it downloads is down, so the tests are failing.

This is what gets downloaded currently as a SEDML model

> <!DOCTYPE html><html><head><title>Apache Tomcat/8.0.24 - Error report</title><style type="text/css">H1 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:22px;} H2 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:16px;} H3 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:14px;} BODY {font-family:Tahoma,Arial,sans-serif;color:black;background-color:white;} B {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;} P {font-family:Tahoma,Arial,sans-serif;background:white;color:black;font-size:12px;}A {color : black;}A.name {color : black;}.line {height: 1px; background-color: #525D76; border: none;}</style> </head><body><h1>HTTP Status 404 - /biomodels-main/download</h1><div class="line"></div><p><b>type</b> Status report</p><p><b>message</b> <u>/biomodels-main/download</u></p><p><b>description</b> <u>The requested resource is not available.</u></p><hr class="line"><h3>Apache Tomcat/8.0.24</h3></body></html>

So, once the site runs fine, these tests will pass fine.

hemilpanchiwala:
For the tons of warnings of [VIOLATION]...., I don't have any idea about them currently as I have not worked on them. I will see that test file and try to figure out why this warnings come.

zakharc:
@hemilpanchiwala: 
> so the tests are failing.

As I said in my previous post (and highlighted in **bold** text): 
I see 5 **errors** (not **fails**) with NPE.
_Fail:_ Our code does not meet its test criteria. That means that the test case fails due to assertions not being fulfilled.
_Error:_ We have an unexpected situation during test execution.
Do we talk about the same tests?

> the reason for Travis CI failing currently is that for the simulation of the SED-ML models, it downloads the models via a URN from biomodels. But currently, the site from which it downloads is down.

I am quite sure, that throwing NPE on this scenario is not the best idea. If this is the reason for the errors, we need a meaningful exception for that. 
> So, once the site runs fine, these tests will pass fine.

I think re-triggering the build would be a good idea. I would like to see it pass :)
> For the tons of warnings of [VIOLATION]...., I don't have any idea about them currently as I have not worked on them. I will see that test file and try to figure out why these warnings come.

:thumbsup:

hemilpanchiwala:
@zakharc,

> As I said in my previous post (and highlighted in bold text):
I see 5 errors (not fails) with NPE.
Fail: Our code does not meet its test criteria. That means that the test case fails due to assertions not being fulfilled.
Error: We have an unexpected situation during test execution.
Do we talk about the same tests?

Yes, I am talking about these tests only. I get the model as a string which I sent in my last message. Then this model is passed to SBMLReader which reads it by giving warnings. And at a point, when any field is accessed from this, it gets a null value. This is the reason why NPE is given.

I have run the tests again but currently, also the tests are failing. But I am sure that the Travis will pass once this thing is running as the [fern](https://github.com/hemilpanchiwala/SBSCL/tree/fern) branch has a passing Travis in my forked repository.

src/main/java/fern/network/sbml/SBMLEventHandlerObserver.java
```diff
@@ -1,128 +1,128 @@
...
-		this.setSimulator(sim);
-	}
-	
-	@Override
-	public String toString() {
-		return name;
-	}
-
-}
+package fern.network.sbml;
```
matthiaskoenig, line 1: I am confused on this file. It says everything changed, but this is clearly code which existed before somewhere ?! Same with the `example/Dsmts.java` above.

src/main/java/fern/simulation/Simulator.java
```diff
@@ -244,6 +244,15 @@ public void initialize() {
 		for (int i=0; i<a.length; i++) 
 			a[i] = propensityCalculator.calculatePropensity(i, getAmountManager(),this);
```
matthiaskoenig, line 245: Complete code duplication! Please no copy paste of code, put this  in a helper function `initializePropensities` (plural!) which is called from initialize. Also missing `{` brackets on for loop!
``` 
		for (int i=0; i<a.length; i++) 
			a[i] = propensityCalculator.calculatePropensity(i, getAmountManager(),this);
``

File: src/main/java/org/simulator/io/CSVImporter.java [hunk 3]
```diff
@@ -153,7 +153,7 @@ public MultiTable convert(Model model, String pathname) throws IOException {
     if (line != null) {
       header = line.split(",");
       line = reader.readLine();
-      while (line != null) {
+      while ((line != null) && !line.isEmpty()) {
```
matthiaskoenig, line 156: Not even sure where to start here. The whole read function has many problems.
First thing: this has as a side effect a change in the header instance variable but at the same time returns a string 2d array. This does not make any sense to me.
This complete module should be cleaned up.

1. One function `readDataFromCSV(String pathname)` which returns a HashMap<String, Double[]>. No global side effects, all columns are read and directly accessible via their key, do the parsing to Doubles right here. No need to generate another hashmap on the outside, no need for finding indices of header keys. This should be a public function which can be reused for other purposes.
2. `convert` should have a better name; i.e. `adaptDataToModel` or something similar and should just take a HashMap<String, Double[]> as second argument. Please document what is exactly happening in the method, i.e. how is the data filtered/adapted, ... to fit to the model.

src/main/java/org/simulator/sbml/SBMLinterpreter.java
```diff
@@ -1220,16 +1195,20 @@ public void init(boolean renewTree, double defaultSpeciesValue, double defaultPa
...
+    int sizeY = model.getCompartmentCount() + model.getSpeciesCount() + model.getParameterCount() + speciesReferencesInRateRules;
+    Y = new double[sizeY];
+    oldY = new double[sizeY];
+    oldY2 = new double[sizeY];
+    changeRate = new double[sizeY];
+    isAmount = new boolean[sizeY];
+    compartmentIndexes = new int[sizeY];
+    conversionFactors = new double[sizeY];
+    inConcentrationValues = new boolean[sizeY];
+    Arrays.fill(conversionFactors, 1d);
```
matthiaskoenig, line 1208: write this directly under the definition of the conversionFactors! don't split initialization from the definition.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -88,9 +87,40 @@ public StochasticTestSuiteTest(String path) {
 
   }
 
-  @Ignore
   @Test
   public void testModel() throws IOException {
+
+    long[] stochasticSeeds = new long[]{
+            1595487468503L, 1595487468503L, 1595487468503L, 1595487468503L, 1595487468503L,
```
matthiaskoenig, line 94: If you use the same seed for everything just do the initialization with a one liner (Arrays.fill as above) instead of defining for every element.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -302,52 +337,49 @@ public void testModel() throws IOException {
...
           }
         }
-      }
 
-    }
-
-
-  }
+        for (Double meanDistance : meanDistances) {
+          Assert.assertTrue((meanDistance > -3d) && (meanDistance < 3d));
```
matthiaskoenig, line 370: Define constants for the hard cutoffs! and document what they mean.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -302,52 +337,49 @@ public void testModel() throws IOException {
...
+        }
 
-  private static GnuPlot runSimulation(Simulator sim, AmountIntervalObserver obs,
-      Map<String, Object> orderedArgs) throws IOException {
+        for (Double sdDistance : sdDistances) {
+          Assert.assertTrue((sdDistance > -5d) && (sdDistance < 5d));
```
matthiaskoenig, line 374: Define constants for hard cut offs

src/main/java/fern/network/sbml/SBMLEventHandlerObserver.java
hemilpanchiwala: @matthiaskoenig, this happened due to different line separators in Windows (CRLF) and Ubuntu (LF). The file content remains the same only the line separators are different (as FERN is developed in Windows and I am working on Ubuntu).

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
hemilpanchiwala: For this, I am using the same seed values for most of the test cases but test cases 28, 29, and 37 are having different seed values. But yes I am updating with Arrays.fill and updating the seed values of left 3 test cases separately.

hemilpanchiwala:
> Please update/refactor the CSVImporter and fix the smaller issues I commented on.

@matthiaskoenig, I have updated all the requested changes.

src/main/java/org/simulator/io/CSVImporter.java
```diff
@@ -58,114 +59,109 @@
...
+              identifierToColumn.get(header[j])[i] = Double.parseDouble(column[j]);
+            }
+          }
+        }
       }
     }
-    reader.close();
-    return result;
+
+    return adaptDataToModel(model, identifierToColumn);
```
matthiaskoenig, line 164: This still mixes all the functions together. 
1. The `readDataFromCSV`should return the HashMap (`Map<String, double[]> identifierToColumn = new HashMap<>();`). Do not call the adaptDataToModel function in the readDataFunction. Also call the hashMap just `columns` or `columnsMap`. Remove the header instance variable and the side effects by it! The header keys are the keys in the HashMap, no need for an instance variable here. You can preserve the order by using a LinkedHashMap (see https://stackoverflow.com/questions/10710193/how-to-preserve-insertion-order-in-hashmap).

2. Add a second function readMultitableFromCSV which returns a MultiTable and uses the readData function. It can have an optional model instance. This is the `adaptDataToModel`. 

Unfortunately I still don't understand what the function is doing exactly. Can you describe this in 1-2 sentences? I am very sure the current implementation is much too complicated for what the function should do. But unfortunately I have no clue what the function should do.

hemilpanchiwala: @matthiaskoenig, I have updated the things you requested. 

In the latest commit, 
readDataFromCSV():
This method takes a path and reads each of the columns from the CSV files and stores them in a LinkedListHashMap with the key as `column name` and value as `an array with its corresponding amounts`.

readMultiTableFromCSV():
In this method, even I am quite confused as I don't see any need of model here. I have taken that model conditions directly from the previous implementation and have copied it but I don't see it of any use (but haven't removed it yet as I am not sure whether it has some use case or not). Other than model usage, I have now taken the map from the readDataFromCSV() method and converted it to the MultiTable.

Also with this, I have removed the global `header` array.

src/main/java/org/simulator/io/CSVImporter.java
```diff
@@ -27,13 +27,7 @@
...
 import java.io.IOException;
-import java.util.AbstractList;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Arrays;
+import java.util.*;
 import java.util.logging.Logger;
```
draeger, line 31: It seems we are not consistent in the use of loggers. Here we are using the on-board logger from Java (which is good because it is always shipped directly with the Java distribution) but in other places we are using the log4j logger (which is also good because it comes with JSBML anyway and it has more features). We should decide which logger to use and stick to that one.

src/main/java/org/simulator/io/CSVImporter.java
```diff
@@ -54,19 +48,63 @@
...
+   * @param pathname The path of the CSV file
+   * @return columnsMap -> LinkedListHashMap
    */
-  private String[] header;
+  private Map<String, double[]> readDataFromCSV(String pathname) throws IOException {
+    Map<String, double[]> columnsMap = new LinkedHashMap<>();
+    BufferedReader reader = new BufferedReader(new FileReader(pathname));
+    String line = reader.readLine();
+    List<String> lines = new LinkedList<String>();
+    String[] identifiers = line.split(",");
```
draeger, line 63: I suggest to create another private method with one further argument, namely the separating character. We would call it with `','` (or possibly as a `String`?) assuming that commas are used as a separator. However, it gives us the flexibility to also use the method with other characters.

src/main/java/org/simulator/io/CSVImporter.java
```diff
@@ -54,19 +48,63 @@
...
+    if (line != null) {
+      line = reader.readLine();
+      while ((line != null) && !line.isEmpty()) {
+        lines.add(line);
+        line = reader.readLine();
+      }
+
+      if (identifiers[0].equalsIgnoreCase("time")){
```
draeger, line 71: This `String` literal `"time"` should be defined somewhere as a `public static final String TIME = "time"`.

src/main/java/org/simulator/io/CSVImporter.java
```diff
@@ -54,19 +48,63 @@
...
+        identifiers[0] = identifiers[0].toLowerCase();
+      }
+      for (String s : identifiers) {
+        columnsMap.put(s, new double[lines.size()]);
+      }
+
+      for (int i = 0; i < lines.size(); i++) {
+        String[] column = lines.get(i).split(",");
```
draeger, line 79: So, if we add another method, then this split `String` will become a variable. As always, when reusing the same value, it is already better to create a variable for it.

src/main/java/org/simulator/io/CSVImporter.java
```diff
@@ -54,19 +48,63 @@
...
+        for (int j = 0; j < column.length; j++) {
+          if ((column[j] != null) && (column[j].length() > 0)) {
+            if (column[j].equalsIgnoreCase("INF")) {
+              columnsMap.get(identifiers[j])[i] = Double.POSITIVE_INFINITY;
+            } else if (column[j].equalsIgnoreCase("-INF")) {
+              columnsMap.get(identifiers[j])[i] = Double.NEGATIVE_INFINITY;
+            } else if (column[j].equalsIgnoreCase("NAN")) {
+              columnsMap.get(identifiers[j])[i] = Double.NaN;
```
draeger, line 87: Better use the existing method [Double.parsedouble](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/lang/Double.html#parseDouble(java.lang.String)) from Java.

src/main/java/org/simulator/sbml/SBMLinterpreter.java
```diff
@@ -1220,16 +1195,20 @@ public void init(boolean renewTree, double defaultSpeciesValue, double defaultPa
...
-      inConcentrationValues = new boolean[Y.length];
-      symbolIdentifiers = new String[Y.length];
-      conversionFactors = new double[Y.length];
-      Arrays.fill(conversionFactors, 1d);
-    }
 
+    int sizeY = model.getCompartmentCount() + model.getSpeciesCount() + model.getParameterCount() + speciesReferencesInRateRules;
+    Y = new double[sizeY];
+    oldY = new double[sizeY];
+    oldY2 = new double[sizeY];
```
draeger, line 1202: Maybe the name `oldY2` is not very intuitive. A comment would be helpful.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -266,7 +262,7 @@ public void testModel() throws IOException {
 
           // Runs the simulation again
           try {
-            runSimulation(sim, obs, orderedArgs);
+            sim.start((Double)orderedArgs.get("time"));
```
draeger, line 371: We really need one central place where the `String` literal `TIME = "time"` should be stored.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -253,6 +253,8 @@ public void testModel() throws IOException {
         // Runs the stochastic simulation repeatedly
         for (int p = 1; p < TOTAL_SIMULATION_COUNT; p++) {
 
+          System.out.println(p);
```
draeger: Please do not use `System.out.print...` but logging.

src/main/java/org/simulator/io/CSVImporter.java
hemilpanchiwala: @draeger, I think it is better to use the log4j logger as it comes from JSBML anyways.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
hemilpanchiwala: We can make an enum class for constants that are used in various classes. What do you think about it?

src/main/java/org/simulator/io/CSVImporter.java
hemilpanchiwala: For string "INF" and "-INF", it will return NumberFormatException. Yeah, but I have updated for the NaN case.

src/main/java/org/simulator/io/CSVImporter.java
hemilpanchiwala: @draeger, do you here mean adding a method like this?

```
private String[] splitString(String s, String separator) {
    return s.split(separator);
}
```

hemilpanchiwala: And with this method, making a variable for the split String like `private static final String SPLIT = ",";`?

src/main/java/org/simulator/sbml/SBMLinterpreter.java
hemilpanchiwala: @draeger, there is a comment for oldY2 at the time when it is first described. You can see it [here](https://github.com/hemilpanchiwala/SBSCL/blob/fern/src/main/java/org/simulator/sbml/EquationSystem.java#L352).

The comment is 

> Array for saving older Y values (when computing delayed values)

hemilpanchiwala: Or better can we rename it to delayedY?

src/main/java/org/simulator/io/CSVImporter.java
draeger: Actually, I was thinking about something like this:
```
private Map<String, double[]> readDataFromCSV(String pathname) throws IOException {
  return readDataFromCSV(pathname, ",");
}
```
So, we would essentially call another private method with an additional argument. Everything is the same, just the method body would be moved to that new method and the separating `String` becomes a variable (to be used at at least two positions).

src/main/java/org/simulator/sbml/SBMLinterpreter.java
draeger: Thanks for clarifying. I think that's good! It becomes clear when looking at the variable within an IDE rather than within a web browser.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
draeger: Yes, `enum`s are good but if it is an enum for just one item it does not make much sense. Also, it is sometimes a bit tricky to use a `String` name from an enum, in particular if it does not match the enum's variable name. For instance, if the enum item is called `TIME` but it should represent the `String` literal `"time"` then you need to add a variable to that enum object and have to retrieve it, for instance like `MyEnum.TIME.getNameValue()` which is much more cumbersome than just using a `public static final String TIME` value. Also, that enum needs a constructor then, etc.

src/main/java/org/simulator/io/CSVImporter.java
```diff
@@ -140,6 +138,10 @@ public MultiTable readMultiTableFromCSV(Model model, String pathname) throws IOE
     return null;
   }
 
+  private String[] splitString(String s, String separator) {
+    return s.split(separator);
+  }
+
```
draeger: Unclear why this method is needed.

src/main/java/org/simulator/io/CSVImporter.java
```diff
@@ -42,10 +45,7 @@
  */
 public class CSVImporter {
 
-  /**
-   * A {@link java.util.logging.Logger} for this class.
-   */
-  private static final transient Logger logger = Logger.getLogger(CSVImporter.class.getName());
```
draeger, line 53: Does it mean we don't have a logger at all anymore in this class? It would have been sufficient to change the import and also to remove `.getName()` when creating the logger if we want to move to log4j.

src/main/java/org/simulator/io/CSVImporter.java
hemilpanchiwala: Sorry for this, I added it by mistake. Thanks for showing it.

src/main/java/org/simulator/io/CSVImporter.java
hemilpanchiwala: It is not needed in this class anymore as there is no logging done. It was just initialized but not used anywhere so I removed it.

src/main/java/org/simulator/io/CSVImporter.java
hemilpanchiwala: Removed in the latest commit!

src/main/java/org/simulator/sbml/SimpleConstraintListener.java
```diff
@@ -48,6 +48,8 @@
    */
   private static final transient Logger logger = Logger.getLogger(SimpleConstraintListener.class.getName());
 
+  boolean printViolationLog = true;
```
draeger, line 51: What happens if there are multiple constrains in a model? Is this listener assigned to exactly one constraint? Or will this switch decide if messages should be printed out for all constraints?

Also, I don't see a mechanism to switch this back on again. What happens if a constraint is, for example, violated from time 2.5 to 6.4 but subsequently not violated until time 12.3, where the condition again indicates a violation? A user would certainly be interested to see messages for each new violation period.

src/main/java/org/simulator/sbml/EquationSystem.java
```diff
@@ -675,9 +680,14 @@ public void init(boolean renewTree, double defaultSpeciesValue, double defaultPa
...
-                        listener.processViolation(evt);
+                    if (model.getConstraint(i).getUserObject(CONSTRAINT_VIOLATION_LOG) == Boolean.FALSE) {
+                        for (ConstraintListener listener: listOfConstraintListeners) {
+                            listener.processViolation(evt);
+                        }
+                        model.getConstraint(i).putUserObject(CONSTRAINT_VIOLATION_LOG, Boolean.TRUE);
                     }
+                } else {
+                    model.getConstraint(i).putUserObject(CONSTRAINT_VIOLATION_LOG, Boolean.FALSE);
                 }
```
draeger: We should modify this again. The `ConstraintEvent` contains an instance of the current `Constraint` object. Therefore, the listener can itself push user objects and check their content. We should have done this right away. To safe the effort for initialization, the `ConstraintListener` implementation could treat a missing user object as `false` and just set it when first needed.

src/main/java/org/simulator/sbml/SimpleConstraintListener.java
```diff
@@ -58,7 +65,22 @@ public void processViolation(ConstraintEvent evt) {
...
     constraint = evt.getSource().getMath().toFormula();
     message = SBMLtools.toXML(evt.getSource().getMessage());
-    // TODO: Localize
-    logger.warn(MessageFormat.format("[VIOLATION]\t{0} at time {1,number}: {2}", constraint, evt.getTime(), message));
+
+    if (evt.getSource().getUserObject(CONSTRAINT_VIOLATION_LOG) == null) {
+      evt.getSource().putUserObject(CONSTRAINT_VIOLATION_LOG, Boolean.FALSE);
+    }
+
+    boolean constraintCondition = ((ASTNodeValue) evt.getSource().getMath().getUserObject(TEMP_VALUE)).compileBoolean(evt.getTime());
```
draeger, line 73: When we reach this line of code, the constraint's condition was causing the listener to become active. Re-evaluation of the condition is unnecessary.

src/main/java/org/simulator/sbml/SimpleConstraintListener.java
```diff
@@ -58,7 +65,22 @@ public void processViolation(ConstraintEvent evt) {
...
+    if ((evt.getSource().getUserObject(CONSTRAINT_VIOLATION_LOG) == Boolean.FALSE) && constraintCondition) {
+      // TODO: Localize
+      logger.warn(MessageFormat.format("[VIOLATION]\t{0} at time {1,number}: {2}", constraint, evt.getTime(), message));
+      evt.getSource().putUserObject(CONSTRAINT_VIOLATION_LOG, Boolean.TRUE);
+    }
+
+    if ((evt.getSource().getUserObject(CONSTRAINT_VIOLATION_LOG) == Boolean.TRUE) && !constraintCondition) {
+      logger.debug(MessageFormat.format("Constraint {0} satisfies again", constraint));
```
draeger, line 81: Little typo: "satisfied again"

draeger:
One more question about the `CSVImporter`: Will it automatically remove blanks between values? If not, this would be an extremely important feature. Maybe instead of using the `equals` method for String comparison, it would be better to check for values using regular expressions with leading and tailing blanks.

hemilpanchiwala:
> One more question about the `CSVImporter`: Will it automatically remove blanks between values? If not, this would be an extremely important feature. Maybe instead of using the `equals` method for String comparison, it would be better to check for values using regular expressions with leading and tailing blanks.

@draeger, I think I can trim each line so spaces get removed and then I can get the things.

src/test/java/org/simulator/fba/BiGGTest.java
```diff
@@ -27,26 +28,30 @@
 public class BiGGTest {
 	private String resource;
     private static final Logger logger = LoggerFactory.getLogger(TestUtils.class);
+    private static final double RESULT_DEVIATION = 1E-6d;
 
+    /**
+     * HashMap for storing the reference results with key as the model names.
```
zakharc, line 34: I doubt that we need to explicitly show, that HashMap will be stored in this field. Also, one can figure out that it's about the reference results from the name of the field.

src/test/java/org/simulator/fba/BiGGTest.java
```diff
@@ -27,26 +28,30 @@
...
-        String value = null;
-        if (env.containsKey(key)) {
-            value = env.get(key);
-            logger.info(String.format("BiGG models folder found: %s", value));
-        }
-        else {
-            logger.info(String.format("%s environment variable not set.", key));
-        }
-        return value;
+        return TestUtils.getPathForTestResource("/bigg/v1.5");
```
zakharc, line 54: Please extract this to a constant.

src/test/java/org/simulator/fba/BiGGTest.java
```diff
@@ -61,36 +66,16 @@ public BiGGTest(String resource) {
 
 		// find all BiGG models (compressed .xml.gz files)
         String biggPath = getBiGGModelPath();
-        System.out.println("BiGG models path: " + biggPath);
-		return TestUtils.findResources(biggPath, ".xml.gz", filter, skip, mvnResource);
+        logger.info("BiGG models path: " + biggPath);
+		return TestUtils.findResources(biggPath, ".xml", filter, skip, mvnResource);
```
zakharc, line 70: This looks awkward. Please adjust this TestUtils method to throw a corresponding exception, e.g. if the resource is missing.

src/test/java/org/simulator/sbml/SBMLTestSuiteRunner.java
```diff
@@ -270,7 +270,7 @@ public static void statisticForSolvers(String file, int from, int to)
 
           File f = new File(csvfile);
           if (f.exists()) {
-            inputData = csvimporter.convert(model, csvfile);
+            inputData = csvimporter.readMultiTableFromCSV(model, csvfile);
```
zakharc, line 273: Camel case.

src/test/java/org/simulator/sbml/SBMLTestSuiteRunner.java
```diff
@@ -474,7 +474,7 @@ else if (modelsWithStrongerTolerance.contains(modelnr)) {
...
```
zakharc, line 477: Same as above.

src/test/java/org/simulator/sbml/SBMLTestSuiteRunner.java
```diff
@@ -718,7 +718,7 @@ public static void testRosenbrockSolverWithSEDML(String file, int from, int to)
             //						mt = createMultiTableFromProcessedResults(
             //								wanted, prRes, sedml);
             CSVImporter csvimporter = new CSVImporter();
-            MultiTable inputData = csvimporter.convert(model,
+            MultiTable inputData = csvimporter.readMultiTableFromCSV(model,
```
zakharc, line 721: Same as above.

src/test/java/org/simulator/sbml/SBMLTestSuiteRunner.java
```diff
@@ -829,7 +829,7 @@ public static MultiTable testModel(AbstractDESSolver solver, Model model,
 
       // solve
       MultiTable solution = solver.solve(interpreter,
-        interpreter.getInitialValues(), timePoints);
+        interpreter.getInitialValues(), timePoints, null);
```
zakharc, line 832: Why is this tabbed?

src/test/java/org/simulator/sbml/SBMLTestSuiteRunner.java
```diff
@@ -954,7 +954,7 @@ public void testModels() throws FileNotFoundException, IOException {
 
           // get timepoints
           CSVImporter csvimporter = new CSVImporter();
-          MultiTable inputData = csvimporter.convert(model, csvfile);
+          MultiTable inputData = csvimporter.readMultiTableFromCSV(model, csvfile);
```
zakharc, line 957: Camel case. 

Please also note the same situation in other objects.

src/test/java/org/simulator/sbml/SBMLTestSuiteRunner.java
```diff
@@ -969,7 +969,7 @@ public void testModels() throws FileNotFoundException, IOException {
             boolean errorInSolve = false;
             try {
               solution = solver.solve(interpreter,
-                interpreter.getInitialValues(), timepoints);
+                interpreter.getInitialValues(), timepoints, null);
```
zakharc, line 972: Please remove the tab.

Please also note the same situation in other objects.

src/test/java/org/simulator/sbml/SBMLTestSuiteRunner.java
```diff
@@ -969,7 +969,7 @@ public void testModels() throws FileNotFoundException, IOException {
...
             } catch (DerivativeException e) {
               errorInSolve = true;
```
zakharc, line 974: Please remove this and throw a "normal" exception. We want to know why the error occurs.

src/test/java/org/simulator/sbml/SBMLTestSuiteTest.java
```diff
@@ -295,7 +295,7 @@ public void testModel() throws IOException {
                         boolean errorInSolve = false;
                         try {
                             solution = solver.solve(interpreter,
-                                    interpreter.getInitialValues(), timepoints);
+                                    interpreter.getInitialValues(), timepoints, null);
                         } catch (DerivativeException e) {
                             errorInSolve = true;
```
zakharc, line 300: Please remove this.

src/test/java/org/simulator/sedml/SEDMLExecutorTest.java
```diff
@@ -181,14 +181,12 @@ public final void testRepeatedScanOscli() throws XMLException, OWLOntologyCreati
     }
 
     @Test
-    @Ignore
     public final void testRepeatedSteadyScanOscli() throws XMLException, OWLOntologyCreationException, IOException {
         String resource = "/sedml/L1V2/repeated-steady-scan-oscli/repeated-steady-scan-oscli.xml";
```
zakharc, line 185: Please extract this to a constant.

src/test/java/org/simulator/sedml/SEDMLExecutorTest.java
```diff
@@ -181,14 +181,12 @@ public final void testRepeatedScanOscli() throws XMLException, OWLOntologyCreati
...
         testSpecificationExample(resource);
     }
 
     @Test
-    @Ignore
     public final void testRepeatedStochasticRuns() throws XMLException, OWLOntologyCreationException, IOException {
         String resource = "/sedml/L1V2/repeated-stochastic-runs/repeated-stochastic-runs.xml";
```
zakharc, line 191: Please extract this to a constant.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -88,9 +198,37 @@ public StochasticTestSuiteTest(String path) {
 
   }
 
-  @Ignore
   @Test
   public void testModel() throws IOException {
+
+    long[] stochasticSeeds = new long[39];
```
zakharc, line 204: Please extract the magic number to a constant.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -88,9 +198,37 @@ public StochasticTestSuiteTest(String path) {
...
+
+    // Initialize the stochasticSeeds array with a same seed value as
+    // most of the test cases pass with same seed.
+    Arrays.fill(stochasticSeeds, 1595487468503L);
```
zakharc, line 208: Please extract the magic number to a constant.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -142,6 +280,11 @@ public void testModel() throws IOException {
         }
         Assert.assertNotNull(sim);
         Assert.assertFalse(errorInSimulator);
+
+        String testcase = path.substring(path.length() - 5);
```
zakharc, line 284: Please add a comment about the magic number.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -142,6 +280,11 @@ public void testModel() throws IOException {
...
+        sim.setStochasticSeed(stochasticSeeds[Integer.parseInt(testcase) - 1]);
+
+        ((SBMLNetwork) net).registerEvents(sim);
         // Initializes the observer for the amounts of molecule species
```
zakharc, line 288: What exactly do we initialize here? I see only an Object reference declaration and definition.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -155,7 +298,7 @@ public void testModel() throws IOException {
         // Runs the stochastic simulation
         boolean errorInSimulation = false;
         try {
-          runSimulation(sim, obs, orderedArgs);
+          sim.start((Double)orderedArgs.get(TIME));
         } catch (Exception e) {
           errorInSimulation = true;
```
zakharc, line 303: Please avoid hiding the actual exceptions.

zakharc, line 303: Same as above.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
```diff
@@ -227,7 +370,7 @@ public void testModel() throws IOException {
 
           // Runs the simulation again
           try {
-            runSimulation(sim, obs, orderedArgs);
+            sim.start((Double)orderedArgs.get(TIME));
           } catch (Exception e) {
             errorInSimulation = true;
           }
```
zakharc, line 376: DRY

src/main/java/fern/simulation/Simulator.java
```diff
@@ -241,9 +241,18 @@ public void initialize() {
 		amountManager.resetAmount();
 		thetaQueue = new ThetaQueue();
 		
-		for (int i=0; i<a.length; i++) 
-			a[i] = propensityCalculator.calculatePropensity(i, getAmountManager(),this);
+		initializePropensities();
 	}
+
+	/**
+	 * Reinitialize the propensities when an event is executed
```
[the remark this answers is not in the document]
hemilpanchiwala: Here, propensities are initialized at first. But while simulating further if any event is executed then they are reinitialized again. Fixing the comment properly.

src/main/java/fern/simulation/Simulator.java
```diff
@@ -367,7 +376,15 @@ public Observer addObserver(Observer observer) {
 		this.observer = n;
 		return observer;
 	}
-	
+
+	/**
+	 * Sets the pre-collected seed for a particular model.
```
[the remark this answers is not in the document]
hemilpanchiwala: As there is randomness in the stochastic tests (due to which Travis can fail), so I have collected the seeds of the test cases when they pass and have set them using this method.
Updating it to Sets the seed for a particular model.

src/main/java/org/simulator/comp/CompSimulator.java
```diff
@@ -112,7 +112,7 @@ public MultiTable solve(double timeEnd, double stepSize, DESSolver solver)
 
     // Compute the numerical solution of the initial value problem
     // TODO: Rel-Tolerance, Abs-Tolerance.
-    MultiTable solution = solver.solve(interpreter, interpreter.getInitialValues(), 0d, timeEnd);
+    MultiTable solution = solver.solve(interpreter, interpreter.getInitialValues(), 0d, timeEnd, null);
```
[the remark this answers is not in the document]
hemilpanchiwala: For this, I will have to make the same copy of the method with just registering the property listener (3-4 lines of code) if any class wants to get the results as they are processed. So, I thought to add null as a parameter to the calls instead of making two methods that are nearly the same. So, should I make a new method with removing the last parameter?

src/main/java/org/simulator/comp/CompSimulator.java
```diff
@@ -112,7 +112,7 @@ public MultiTable solve(double timeEnd, double stepSize, DESSolver solver)
 
     // Compute the numerical solution of the initial value problem
```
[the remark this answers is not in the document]
hemilpanchiwala: Yes, removing it.

src/main/java/org/simulator/sedml/SedMLSBMLSimulatorExecutor.java
```diff
@@ -508,8 +508,8 @@ protected IRawSedmlSimulationResults executeSimulation(String modelStr, Simulati
   /**
    * Merge two 2D arrays into one 2D array in X-direction
    *
-   * @param double[][]
-   * @param double[][]
+   * @param a
+   * @param b
```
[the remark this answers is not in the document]
hemilpanchiwala: Don't have an idea about this!

src/main/java/org/simulator/sedml/SedMLSBMLSimulatorExecutor.java
```diff
@@ -522,8 +522,8 @@ protected IRawSedmlSimulationResults executeSimulation(String modelStr, Simulati
   /**
    * Merge time columns from 2 multiTables
    *
-   * @param MultTableSEDMLWrapper
-   * @param MultTableSEDMLWrapper
+   * @param a
+   * @param b
```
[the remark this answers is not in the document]
hemilpanchiwala: Don't have an idea about this!

src/test/java/org/simulator/fba/BiGGTest.java
hemilpanchiwala: Here, I think there is no exception. Whenever there are no files with this extension, it sends empty Iterable object.

src/test/java/org/simulator/sbml/SBMLTestSuiteRunner.java
hemilpanchiwala: Here, do you suggest to camel case the `csvimporter` to `csvImporter`?

src/test/java/org/simulator/sedml/SEDMLExecutorTest.java
hemilpanchiwala: Here, the resources are provided for all the tests separately (in their corresponding test methods) so, I don't this to make this as a constant.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
hemilpanchiwala: Here, I am making the instance of the observer which keeps track of amounts and others. So added a comment `Initializes the observer ....`.

hemilpanchiwala:
@zakharc, I have updated most of the requested changes. I have some questions in few which I have commented, once you answer that I will update that changes too.

src/test/java/org/simulator/sbml/SBMLTestSuiteRunner.java
hemilpanchiwala: I am not understanding what are you meaning by tab?

src/main/java/org/simulator/comp/CompSimulator.java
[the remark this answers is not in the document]
zakharc: > So, should I make a new method with removing the last parameter?
Yes, please add an overloaded method.

Overloading is a common practice. Some benefits:
- The main advantage of this is the cleanliness of code.
- Method overloading increases the readability of the program.
- Overloaded methods give programmers the flexibility to call a similar method for different types of data.

src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java
zakharc: > AmountIntervalObserver obs = null;
Where exactly are we making the instance of the observer here?

hemilpanchiwala: https://github.com/hemilpanchiwala/SBSCL/blob/fern/src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java#L299

zakharc: In the line, you pointed out, no initialization happens. Only object reference declaration and definition.

The instance of the observer is created here:
https://github.com/draeger-lab/SBSCL/blob/41fb960ace6c6a34b99c4dcb307244c63b1a497d/src/test/java/org/simulator/stochastic/StochasticTestSuiteTest.java#L494

Probably this might help:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.9

hemilpanchiwala: I thought that you wanted the line where the `AmountIntervalObserver obs = null;` was there. Removing the comment from there.

shalinshah1993:
LGTM! :D

src/main/java/org/simulator/sedml/SedMLSBMLSimulatorExecutor.java
```diff
@@ -522,8 +522,8 @@ protected IRawSedmlSimulationResults executeSimulation(String modelStr, Simulati
...
    * @return double[]
    */
   private double[] mergeTimeCols(MultTableSEDMLWrapper a, MultTableSEDMLWrapper b) {
```
shalinshah1993, line 529: should probably name params better than a and b

src/test/java/org/simulator/distance/MaxAbsDistanceTest.java
```diff
@@ -73,8 +73,8 @@ public void testMaxAbsDistance() throws IOException {
 
         // convert the test files to MultiTable
         CSVImporter csvImporter = new CSVImporter();
-        MultiTable a = csvImporter.convert(null, first);
-        MultiTable b = csvImporter.convert(null, second);
+        MultiTable a = csvImporter.readMultiTableFromCSV(null, first);
```
shalinshah1993, line 76: name this variable appropriately instead of a and b

src/test/java/org/simulator/distance/MaxDivergenceToleranceTest.java
```diff
@@ -76,8 +76,8 @@ public void testMaxDivergenceTolerance() throws IOException {
 
         // convert the test files to MultiTable
         CSVImporter csvImporter = new CSVImporter();
-        MultiTable a = csvImporter.convert(null, first);
-        MultiTable b = csvImporter.convert(null, second);
+        MultiTable a = csvImporter.readMultiTableFromCSV(null, first);
```
shalinshah1993, line 79: same as above

src/test/java/org/simulator/sbml/SBMLTestSuiteTest.java
```diff
@@ -181,7 +167,7 @@ public void testModel() throws IOException {
 
                 // get timepoints
                 CSVImporter csvimporter = new CSVImporter();
```
shalinshah1993, line 169: camel case

src/test/java/org/simulator/sbml/SBMLTestSuiteWrapper.java
```diff
@@ -161,7 +161,7 @@ else if (modelsWithStrongerTolerance.contains(modelnr)) {
     }
     if (model != null) {
       CSVImporter csvimporter = new CSVImporter();
-      MultiTable inputData = csvimporter.convert(model, csvfile);
+      MultiTable inputData = csvimporter.readMultiTableFromCSV(model, csvfile);
```
shalinshah1993, line 164: csvImporter

src/main/java/org/simulator/fba/FluxBalanceAnalysis.java
```diff
@@ -235,7 +238,7 @@ public FluxBalanceAnalysis(SBMLDocument doc)
         for (Pair<String, Double> pair : species2Reaction.get(species.getId())) {
           weights[reaction2Index.get(pair.getKey())] = pair.getValue();
         }
-        if (species2Reaction.get(species.getId()).size() > 1) {
+        if (species.isSetBoundaryCondition() && !species.getBoundaryCondition()) {
           problem.addConstraint(new LinearEqualsConstraint(weights, 0.0, "cnstrt_" + species.getId()));
```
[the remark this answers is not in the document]
hemilpanchiwala: @draeger, Updated in the latest commit!

PR #59, opened 2021-01-10
Fix ErrorProne issues and Bash download script

curiousleo:
Hi! I ran [ErrorProne](http://errorprone.info) over this project and found a few issues -- documented inline.

The docs in `src/docs` seem out of date: after `mvn install` there are lots of changes. I excluded those changes from this PR to not make it too big.


src/main/java/org/simulator/sbml/EquationSystem.java
```diff
@@ -1809,9 +1809,8 @@ public void propertyChange(PropertyChangeEvent propertyChangeEvent) {
    * @param highOrderEvents
    */
   protected void pickRandomEvent(List<Integer> highOrderEvents) {
-    int length = highOrderEvents.size();
-    int random = RNG.randomInt(0, length - 1);
```
curiousleo, line 1813: `randomInt` was the only method in `RNG` that was used anywhere.

`Random.nextInt(n)` does the same thing as `RNG.nextInt(0, n-1)`, so we can just use the `Random` API directly here.

`ThreadLocalRandom.current()` is simply a fast `Random` instance.

src/main/java/org/simulator/sbml/SBMLinterpreter.java
```diff
@@ -107,7 +107,7 @@ public SBMLinterpreter(Model model)
   public SBMLinterpreter(Model model, double defaultSpeciesValue, double defaultParameterValue,
       double defaultCompartmentValue)
       throws SBMLException, ModelOverdeterminedException {
-    this(model, 0d, 1d, 1d, null);
```
curiousleo, line 110: ErrorProne complained about this -- the constructor arguments were previously ignored, which looks unintended.

src/main/java/org/simulator/sbml/SBMLinterpreter.java
```diff
@@ -548,9 +548,7 @@ private SBMLEventInProgress processNextEvent(HashSet<Double> priorities, double[
         pickRandomEvent(highOrderEvents);
       }
     } else {
-      for (int i = 0; i < runningEvents.size(); i++) {
-        highOrderEvents.add(runningEvents.get(i));
-      }
+      highOrderEvents.addAll(runningEvents);
```
curiousleo, line 551: This just seemed both more readable and potentially faster.

src/test/java/org/simulator/oven/SCPSolverIssue.java
```diff
@@ -17,11 +18,11 @@ public static void main(String[] args) {
 
     LinearProgramSolver solver = SolverFactory.newDefault();
     double[] sol = solver.solve(lp);
-    System.out.println(sol);
```
curiousleo, line 20: ErrorProne complained about this: printing an array directly does not show anything useful.

src/test/scripts/download_bigg_models.sh
```diff
@@ -6,6 +6,8 @@
 #  source ./download_bigg_models.sh
 #################################################
 
+set -euo pipefail
```
curiousleo, line 9: Stop the script if there is an error.

src/test/scripts/download_sbml-test-suite.sh
```diff
@@ -6,14 +6,15 @@
...
+set -euo pipefail
+
 # This URL gives the updated SBML test cases from SBML Test Suite repository
 SBML_TEST_SUITE_LINK="https://github.com/sbmlteam/sbml-test-suite/branches/develop/cases"
 
 _CWD="$PWD"
 TEST_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 # download and extract the bigg models for testing
-cd $TEST_DIR/../resources/sbml-test-suite
```
curiousleo, line 16: This looks like a copy-paste mistake: it only makes sense to `cd` after the `rm` and `mkdir` (see following lines).

shalinshah1993:
Thank you for the wonderful edits @curiousleo

PR #60, opened 2021-01-11
Fewer BigDecimal conversions

curiousleo:
I took a look at the tests with [`perf`](https://perf.wiki.kernel.org/index.php/Main_Page) and saw that a bunch of time was spent in `BigDecimal` conversions. At least those in the hot loop seem unnecessary, so I removed them.

### DIFF
--- a/src/main/java/org/simulator/math/odes/RosenbrockSolver.java
+++ b/src/main/java/org/simulator/math/odes/RosenbrockSolver.java
@@ -264,6 +264,17 @@ public class RosenbrockSolver extends AdaptiveStepsizeIntegrator {
     JAC = new double[numEqn][numEqn];
     FAC = new double[numEqn][numEqn];
     I = new double[numEqn][numEqn];
+
+    for (int i = 0; i < numEqn; i++) {
+      for (int j = 0; j < numEqn; j++) {
+        if (i == j) {
+          I[i][j] = 1;
+        } else {
+          I[i][j] = 0;
+        }
+      }
+    }
+
     ignoreNaN = new boolean[numEqn];
   }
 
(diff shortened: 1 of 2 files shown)


matthiaskoenig (approved):
Changes look great to me. Much more readable and clearer then before.

hemilpanchiwala (approved):
Changes look good to me.

draeger (approved):

draeger:
@curiousleo: Travis reports that the build fails after the changes to the download script.

src/main/java/org/simulator/math/odes/RosenbrockSolver.java
```diff
@@ -296,15 +307,6 @@ public double step(DESystem DES) throws DerivativeException {
...
     }
-    for (int i = 0; i < numEqn; i++) {
-      for (int j = 0; j < numEqn; j++) {
-        if (i == j) {
-          I[i][j] = 1;
-        } else {
-          I[i][j] = 0;
-        }
-      }
-    }
```
curiousleo, line 307: Moved to `init()`, no need to initialise the identity matrix anew in every `step()`.

src/main/java/org/simulator/math/odes/RosenbrockSolver.java
```diff
@@ -440,8 +442,7 @@ public int getNumEquations() {
         hasDerivatives = false;
       }
     }
-    double timeEnd = BigDecimal.valueOf(time).add(BigDecimal.valueOf(currentStepSize))
-        .doubleValue();
+    double timeEnd = time + currentStepSize;
```
curiousleo, line 445: According to the IEEE754 spec, floating point addition must already be "as precise as possible", i.e. return the closest floating point number to the real (in the sense of real numbers) result. So I don't think there is any benefit in translation to `BigDecimal` and back.

src/main/java/org/simulator/math/odes/RosenbrockSolver.java
```diff
@@ -475,11 +476,7 @@ public int getNumEquations() {
         System.arraycopy(y2, 0, y, 0, y.length);
       }
       for (int i = 0; i != y.length; i++) {
-        if (Double.isInfinite(y[i]) || (Double.isNaN(y[i]))) {
-          ignoreNaN[i] = true;
-        } else {
-          ignoreNaN[i] = false;
-        }
+        ignoreNaN[i] = !Double.isFinite(y[i]);
```
curiousleo, line 479: `!isFinite(x) <==> isInfinite(x) \/ isNaN(x)`

curiousleo:
> @curiousleo: Travis reports that the build fails after the changes to the download script.

Oops, I had no intention of breaking your build!

- [`master` is green](https://travis-ci.com/github/draeger-lab/SBSCL/builds/212750712) after #59 was merged
- [this PR's branch is green](https://travis-ci.com/github/draeger-lab/SBSCL/builds/212883366) (and CI runs about a minute faster!)

Could you point me to the failure you're seeing after the download script changes?

PR #61, opened 2021-01-13
Refactors

curiousleo:
A few automated refactors. The commits are self-contained.

Feel free to ignore - some of these are stylistic questions about which opinions can differ!

### DIFF
--- a/src/main/java/fern/analysis/NetworkSearchAction.java
+++ b/src/main/java/fern/analysis/NetworkSearchAction.java
@@ -20,28 +20,28 @@ public interface NetworkSearchAction {
    *
    * @param reaction index of the inserted reaction
    */
-  public void reactionDiscovered(int reaction);
+  void reactionDiscovered(int reaction);
 
   /**
    * Gets called when a species is inserted into the search structure.
    *
    * @param species index of the inserted species
    */
-  public void speciesDiscovered(int species);
+  void speciesDiscovered(int species);
 
   /**
    * Gets called when a reaction gets out of the search structure.
    *
    * @param reaction index of the reactions
    */
-  public void reactionFinished(int reaction);
+  void reactionFinished(int reaction);
 
   /**
    * Gets called when a species gets out of the search structure.
    *
    * @param species index of the species
    */
-  public void speciesFinished(int species);
+  void speciesFinished(int species);
 
   /**
    * Gets called, before the species is inserted into the search structure. If the implementing
@@ -51,7 +51,7 @@ public interface NetworkSearchAction {
    * @param neighborType one of the NeighborTypes
    * @return true if the species should be inserted
    */
-  public boolean checkSpecies(int species, NeighborType neighborType);
+  boolean checkSpecies(int species, NeighborType neighborType);
 
   /**
    * Gets called, before the reaction is inserted into the search structure. If the implementing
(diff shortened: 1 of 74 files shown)


draeger:
My only concern with the suggested changes would be if they enforce an upgrade in the end-users' Java™ version? We need to carefully consider if we loose users of SBSCL by enforcing higher Java versions or if it is backward compatible? Which would be the minimal Java version that can still be used after these changes?

src/main/java/fern/analysis/NetworkSearchAction.java
```diff
@@ -20,28 +20,28 @@
    *
    * @param reaction index of the inserted reaction
    */
-  public void reactionDiscovered(int reaction);
```
shalinshah1993, line 23: What's the argument for deleting `public` here? It is nice if we have explicit mention.

src/main/java/fern/analysis/ShortestPath.java
```diff
@@ -202,7 +202,7 @@ public ShortestPathAction(NodeChecker checker) {
      * @return
      */
     public Path getPath(int species) {
-      LinkedList<Integer> path = new LinkedList<Integer>();
```
shalinshah1993, line 205: I like the explicit mention type when initializing. Makes things little easy to follow. I'd prefer keeping this the original way.

draeger (commented):
@curiousleo: My only concern with the suggested changes would be if they enforce an upgrade in the end-users' Java™ version? We need to carefully consider if we loose users of SBSCL by enforcing higher Java versions or if it is backward compatible? Which would be the minimal Java version that can still be used after these changes?

src/main/java/fern/analysis/NetworkSearchAction.java
curiousleo: https://docs.oracle.com/javase/tutorial/java/IandI/interfaceDef.html

> All abstract, default, and static methods in an interface are implicitly public, so you can omit the public modifier.

> All constant values defined in an interface are implicitly public, static, and final. Once again, you can omit these modifiers.

Declaring an interface method `public` carries the same amount of information as `/* this is a comment */` :)

src/main/java/fern/analysis/ShortestPath.java
curiousleo: Fair! I personally find Java verbose enough with all the _necessary_ declaration, so I tend to trim the unnecessary ones as much as possible - it seems clear here even in the less explicitly typed version that a `LinkedList` containing `Integer`s is being created. But tastes differ!

curiousleo:
> @curiousleo: My only concern with the suggested changes would be if they enforce an upgrade in the end-users' Java™ version? We need to carefully consider if we loose users of SBSCL by enforcing higher Java versions or if it is backward compatible? Which would be the minimal Java version that can still be used after these changes?

Fair point!

Changing `jdk.version` in `pom.xml` and then running `mvn compile` shows that `master` as of 7d9548820719b587f36c54157537700afd40dbec

- Fails to build against JDK 6: "diamond operator is not supported"
- Fails to build against JDK 7: "lambda expressions are not supported"
- Successfully builds against JDK 8 (that's what the default is in `pom.xml` and what runs in CI)

This PR also builds against JDK 8 (if not, it would have failed CI). So there's no regression in terms of which JDK versions SBSCL can build against, compatibility is unchanged.

draeger:
> > @curiousleo: My only concern with the suggested changes would be if they enforce an upgrade in the end-users' Java™ version? We need to carefully consider if we loose users of SBSCL by enforcing higher Java versions or if it is backward compatible? Which would be the minimal Java version that can still be used after these changes?
> 
> Fair point!
> 
> Changing `jdk.version` in `pom.xml` and then running `mvn compile` shows that `master` as of [7d95488](https://github.com/draeger-lab/SBSCL/commit/7d9548820719b587f36c54157537700afd40dbec)
> 
> * Fails to build against JDK 6: "diamond operator is not supported"
> * Fails to build against JDK 7: "lambda expressions are not supported"
> * Successfully builds against JDK 8 (that's what the default is in `pom.xml` and what runs in CI)
> 
> This PR also builds against JDK 8 (if not, it would have failed CI). So there's no regression in terms of which JDK versions SBSCL can build against, compatibility is unchanged.

Ok, I think, Java 8 is nearly 7 years old and, therefore, a requirement we can assume users to have installed by now.

PR #64, opened 2021-02-22
Updates in the BiGG models test

hemilpanchiwala:
This PR updates the following:
- Updates the script to download the newer version of BiGG models
- Adds functionality to add the results of tests in the csv file
- Reference results of Cobrapy are also updated to the newer version of BiGG database


draeger (approved):
Great job!

src/test/scripts/download_bigg_models.sh
```diff
@@ -9,22 +9,21 @@
 set -eo pipefail
 
 # This URL gives the updated BiGG Models from the repository
-BIGG_MODELS_BASE_URL="https://github.com/matthiaskoenig/bigg-models-fba/raw/master/models"
-FILE_NAME="bigg_models_v1.5.tar.gz"
+BIGG_MODELS_BASE_URL="https://www.dropbox.com/sh/ye05djxrpxy37da/AAAPwL27xQ97mHGoLmCxhI-4a/v1.6"
```
shalinshah1993, line 12: whose dropbox is this? It might be tricky if this is your's hemil incase if might be gone in future. We can have @draeger host this file on his dropbox.

shalinshah1993 (approved):
Left a small comment about dropbox URL but otherwise LGTM! 🚀 
Thanks for the help @hemilpanchiwala

src/test/scripts/download_bigg_models.sh
hemilpanchiwala: @shalinshah1993, this link is from the official BiGG models repository only. They store the versions of BiGG database in dropbox. You can check it [here](https://github.com/SBRG/bigg_models#dumping-and-restoring-the-database).

PR #67, opened 2021-04-22
BioModels support

hemilpanchiwala:
- Adds script to download BioModels
- Adds test for the BioModels
- Adds support for saving the simulation results of stochastic models in the CSV format


draeger:
Thanks, @hemilpanchiwala! Great work. I am just a bit confused about the changes to the old submission to BMC Bioinformatics and don't quite understand why the TeX files had to be changed? Also, it seems a lot of BioModels are being pushed to the repository, even though the `.gitignore` file states they should be excluded. Also there seems to be something wrong with the CSV files for the stochastic tests because the `.gitignore` file tries to exclude their containing folder but the path has an error so that they are included anyway. Do you think, we could exclude all those resource files effectively and also keep the older TeX files unchanged?

hemilpanchiwala:
@draeger, Yes the biomodels were included in the commit even after I added them in .gitignore. So, I had to fix this and all biomodels are removed after the `.gitignore fix` commit.

For the stochastic models, I have not added the results in the .gitignore files as they can be present in the repository for reference. So, they are knowingly left in the repository.

The .tex files are changed also due to the .gitignore fix only but they are left unchanged (i.e. whatever is removed is added back). I followed this for updating the repository according to .gitignore (https://www.codeblocq.com/2016/01/Untrack-files-already-added-to-git-repository-based-on-gitignore/).

draeger (commented):
Looks good. I'd like to merge if it is clear that certain files are skipped (as discussed before).

hemilpanchiwala:
@draeger, I think you can merge as I have removed all the BioModels from the repository.

For the results of the stochastic test cases, I think we should keep them in the repository for reference and even other softwares can compare their results with ours if they want directly from the link (which would be quite handy) instead of simulating the models in his/her local computer after installing SBSCL.

PR #72, opened 2023-07-07
Match the version of JSBML that we have in maven central.

funasoul:
Match the version of JSBML (1.6.1) located at maven central. Also, the INSTALL.md example assumes the bash user, so it has been changed to work with other shell users (zsh, etc.).


draeger (approved):

draeger (approved):

draeger:
Thank you, @funasoul, for these corrections!

PR #75, opened 2025-03-31
Fix daxpy_InPlaceModification Test, Ddot_NaN Test, and Add new Ddot_L…

Aditya201003:
1. **Fixed NaN Check in Ddot_NaN() (Line 117)**  
   - **Problem:** Used `assertEquals(Double.NaN, result, 1e-6)`, but in Java, `NaN != NaN`, causing the test to fail even when correct.  
   - **Fix:** Replaced with `assertTrue(Double.isNaN(result));`, ensuring proper NaN verification.  

2. **Modified daxpy_InPlaceModification() Test (Line 822)**  
   - **Problem:** `dy = dx;` caused `dy` to reference the same memory as `dx`, leading to unintended modifications.  
   - **Fix:** Created a separate copy of `dx` for `dy` to prevent in-place modification.  

3. **Added Ddot_LargeNumbers() Test (Line 792)**  
   - **New Test:** Checks `ddot` with extremely large and small floating-point numbers.  
   - **Placement:** Added above the existing `daxpy_LargeValues()` test for better test coverage.  
   - **Purpose:** Ensures `ddot` correctly handles edge cases involving precision and floating-point limitations.

### DIFF
--- a/src/test/java/LSODAIntegratorTest.java
+++ b/src/test/java/LSODAIntegratorTest.java
@@ -114,8 +114,9 @@ public class LSODAIntegratorTest {
         int incx = 1; int incy = 1;
 
         double result = LSODAIntegrator.ddot(n, dx, dy, incx, incy);
-        assertEquals(Double.NaN, result, 1e-6);
+        assertTrue(Double.isNaN(result));
     }
+
 
     
     /** 
@@ -788,6 +789,17 @@ public class LSODAIntegratorTest {
         assertArrayEquals(expectedDy, dy, 1e-6);
     }
 
+     @Test
+    void Ddot_LargeNumbers() {
+        int n = 2;
+        double[] dx = {0d, 1e100, -1e100};
+        double[] dy = {0d, 1e-100, 1e-100};
+        int incx = 1, incy = 1;
+
+        double result = LSODAIntegrator.ddot(n, dx, dy, incx, incy);
+        assertEquals(1e100 * 1e-100 + (-1e100) * 1e-100, result, 1e-6);
+    }
+
     @Test
     void daxpy_LargeValues() {
         int n = 3;
@@ -807,7 +819,7 @@ public class LSODAIntegratorTest {
         int n = 3;
         double da = 2d;
         double[] dx = {0d, 1d, 2d, 3d};
-        double[] dy = dx;
+        double[] dy = dx.clone();
         int incx = 1, incy = 1;
 
         LSODAIntegrator.daxpy(n, da, dx, incx, incy, dy);
@@ -815,6 +827,7 @@ public class LSODAIntegratorTest {
         double[] expectedDy = {0d, 3d, 6d, 9d};
         assertArrayEquals(expectedDy, dy, 1e-6);
     }
+
 
     /**
      * The following tests are for the functions dgefa() within LSODAIntegrator
--- a/src/test/java/LSODAIntegratorTest.java
+++ b/src/test/java/LSODAIntegratorTest.java
@@ -107,7 +107,7 @@ public class LSODAIntegratorTest {
     }
 
     @Test
-    void ddot_NaN() {
+    void ddotNaN() {
         int n = 2;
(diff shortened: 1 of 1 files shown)


TlexCypher:
@Aditya201003 
Thanks for great contributions.
I put some nits comments, I hope that these suggested changes might be helpful for you.

src/test/java/LSODAIntegratorTest.java
```diff
@@ -788,6 +789,17 @@ void daxpy_AllZeroDx() {
         assertArrayEquals(expectedDy, dy, 1e-6);
     }
 
+     @Test
+    void Ddot_LargeNumbers() {
```
TlexCypher, line 793: As well as @NeumannArthur, in Java, snake case is not following conventions.
As same as here, all functions and variable names should be changed to camel case.
If you want to know about Java conventions, you should explore major linters and build CI/CD pipelines to protect code quality.

By the way, I prefer this [Linter](https://github.com/nebula-plugins/gradle-lint-plugin), built by Netflix.
But this linter works under only gradle project.
I created [new issues](https://github.com/draeger-lab/SBSCL/issues/76, https://github.com/draeger-lab/SBSCL/issues/77) to introduce linters, so if you want to contribute more, work on it.
I'll help you with my best effort.

src/test/java/LSODAIntegratorTest.java
```diff
@@ -788,6 +789,17 @@ void daxpy_AllZeroDx() {
...
+        int n = 2;
+        double[] dx = {0d, 1e100, -1e100};
+        double[] dy = {0d, 1e-100, 1e-100};
+        int incx = 1, incy = 1;
+
+        double result = LSODAIntegrator.ddot(n, dx, dy, incx, incy);
+        assertEquals(1e100 * 1e-100 + (-1e100) * 1e-100, result, 1e-6);
```
TlexCypher, line 800: [nits]
Why is the expected value written as an expression like 1e100 * 1e-100 + (-1e100) * 1e-100?
You should express what it means using variable names to make the intention clear.

TlexCypher (changes requested):

TlexCypher:
@draeger 
Basically, we need the branch protection rule.
Could you add some rules.
In my opinion, at least two reviewers must put LGTM and contributers should resolve all request changes.
After building CI/CD pipelines, all tests and all lint and format check should be passed.
How do you think?

src/test/java/LSODAIntegratorTest.java
```diff
@@ -107,14 +107,14 @@ void Ddot_SingleElement() {
     }
 
     @Test
-    void Ddot_NaN() {
+    void ddot_NaN() {
```
TlexCypher, line 110: Suggested changes.
```java
void ddotNan() {
    /*body of test*/
}
```

NeumannArthur (approved):

src/test/java/LSODAIntegratorTest.java
```diff
@@ -788,6 +788,20 @@ void daxpy_AllZeroDx() {
...
+    
+        double sumOfProducts = (dx[1] * dy[1]) + (dx[2] * dy[2]); 
+        double result = LSODAIntegrator.ddot(n, dx, dy, incX, incY);
+    
+        assertEquals(sumOfProducts, result, 1e-6);
+    }
+
+
     @Test
     void daxpy_LargeValues() {
```
TlexCypher, line 806: Could you change to camelCase?

TlexCypher (changes requested):
Could you add some assetions on failed testcase?
It is only written passed case.

TlexCypher:
@Aditya201003 I created another issue related LSODAIntegrator.java.
Could you work on it in parallel as well?
Your contribution is so great, so I want you work on it.

TlexCypher:
@draeger @NeumannArthur, please don't merge this PR until I give LGTM.

TlexCypher (approved):

PR #78, opened 2025-04-03
Add support for RK2 method and minor improvements in SBMLTestSuiteRunnerWrapper and RK4 method

baranwalayush:
### Added Support for RK2 (Heun’s Method)
The library previously lacked RK2 support. While RK4 is more accurate, RK2 provides a useful alternative with fewer computations.
- **Implemented RK2 method using Heun’s method**.
- **User can select between RK2 and RK4 by passing an additional method argument**
- **RK4 remains the default method**.

### Minor improvements in RK4 Solver
- **Precomputed `tMid`** .
- **Optimized `svvAddScaled` calls**: Instead of multiple calls, the final sum is now computed in a single loop.

### Improvements in `SBMLTestSuiteRunnerWrapper.java`
- **Optimized `resultColumns`**: Replaced `HashMap` with `HashSet` as it is only used for checking value presence, reducing memory overhead.
- **Removed redundant condition**: Eliminated an unnecessary `if` condition in the `variablesToAdd` array.

### TODO
- Update documentation for `RungeKutta_EventSolver`.

### DIFF
--- a/src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
+++ b/src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
@@ -26,6 +26,9 @@ package org.simulator.math.odes;
 
 import org.apache.commons.math.ode.DerivativeException;
 import org.simulator.math.Mathematics;
+import java.util.logging.Logger;
+import java.util.Set;
+import java.util.logging.Level;
 
 /**
  * Runge-Kutta method.
@@ -42,6 +45,16 @@ public class RungeKutta_EventSolver extends AbstractDESSolver {
   private static final long serialVersionUID = -2034495479346567501L;
 
   /**
+   * Logger for this class
+   */
+  private static final Logger logger = Logger.getLogger(RungeKutta_EventSolver.class.getName());
+
+  /**
+   * Set of supported RungeKutta methods
+   */
+  private static final Set<String> validMethods = Set.of("RK2", "RK4");
+  
+  /**
    * Stores temporary results for the fourth-order Runge-Kutta method.
    */
   transient protected double[][] kVals = null;
@@ -51,11 +64,30 @@ public class RungeKutta_EventSolver extends AbstractDESSolver {
    */
   transient protected double[] kHelp;
 
+  /** 
+   * String to store the Runge_Kutta Method
+   */
+  private final String method;
+
   /**
    * default constructor
    */
   public RungeKutta_EventSolver() {
     super();
+    this.method = "RK4";
+  }
+
+  /**
+   * @param method
+   */
+  public RungeKutta_EventSolver(String method) {
+    super();
+    if (!validMethods.contains(method)) {
+      logger.log(Level.WARNING, "Unsupported method: {0}. Defaulting to 'RK4'.", method);
+      this.method = "RK4";
+    } else {
+      this.method = method;
+    }
   }
 
   /**
@@ -63,6 +95,21 @@ public class RungeKutta_EventSolver extends AbstractDESSolver {
    */
   public RungeKutta_EventSolver(double stepSize) {
     super(stepSize);
+    this.method = "RK4";
+  }
+
+  /**
+   * @param stepSize
+   * @param method
+   */
+  public RungeKutta_EventSolver(double stepSize, String method) {
+    super(stepSize);
+    if (!validMethods.contains(method)) {
+      logger.log(Level.WARNING, "Unsupported method: {0}. Defaulting to 'RK4'.", method);
+      this.method = "RK4";
+    } else {
+      this.method = method;
+    }
   }
 
   /**
@@ -72,6 +119,23 @@ public class RungeKutta_EventSolver extends AbstractDESSolver {
    */
   public RungeKutta_EventSolver(double stepSize, boolean nonnegative) {
     super(stepSize, nonnegative);
+    this.method = "RK4";
+  }
+
+  /**
+   * @param stepSize
+   * @param nonnegative the nonnegative flag of the super class
+   * @param method
+   * @see AbstractDESSolver
+   */
+  public RungeKutta_EventSolver(double stepSize, boolean nonnegative, String method) {
+    super(stepSize, nonnegative);
+    if (!validMethods.contains(method)) {
+      logger.log(Level.WARNING, "Unsupported method: {0}. Defaulting to 'RK4'.", method);
+      this.method = "RK4";
+    } else {
+      this.method = method;
+    }
   }
 
   /**
@@ -81,6 +145,7 @@ public class RungeKutta_EventSolver extends AbstractDESSolver {
    */
   public RungeKutta_EventSolver(RungeKutta_EventSolver rkEventSolver) {
     super(rkEventSolver);
+    this.method = rkEventSolver.method;
   }
 
   /* (non-Javadoc)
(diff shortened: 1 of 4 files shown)


src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
```diff
@@ -41,6 +44,16 @@ public class RungeKutta_EventSolver extends AbstractDESSolver {
...
 
+  /**
+   * Logger for this class
+   */
+  private static final Logger logger = Logger.getLogger(RungeKutta_EventSolver.class.getName());
+
+  /**
+   * Set of supported RungeKutta methods
+   */
+  private static final Set<String> validMethods = Set.of("RK2", "RK4");
```
TlexCypher, line 55: Imo, You should not use hard-codeded strings.
I prefer you to declare some Enums to handle methods name.

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
```diff
@@ -51,18 +64,52 @@ public class RungeKutta_EventSolver extends AbstractDESSolver {
...
+    this.method = "RK4";
+  }
+
+  /**
+   * @param method
+   */
+  public RungeKutta_EventSolver(String method) {
+    super();
+    if (!validMethods.contains(method)) {
+      logger.log(Level.WARNING, "Unsupported method: {0}. Defaulting to 'RK4'.", method);
```
TlexCypher, line 86: Imo, the action , switching to default, is not categorized as warning level.
I think proper level is info, but what do you think?

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
```diff
@@ -90,40 +155,87 @@ public RungeKutta_EventSolver(RungeKutta_EventSolver rkEventSolver) {
   public double[] computeChange(DESystem DES, double[] yTemp, double t, double h, double[] change,
       boolean steadyState)
       throws DerivativeException {
-    int dim = DES.getDimension();
-    if ((kVals == null) || (kVals.length != 4) || (kVals[0].length != dim)) {
-      // "static" vectors which are allocated only once
-      kVals = new double[4][dim];
-      kHelp = new double[dim];
+
+        if(method.equals("RK2")) {
```
TlexCypher, line 159: Prefer to use switch statement with Enums.

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
```diff
@@ -90,40 +155,87 @@ public RungeKutta_EventSolver(RungeKutta_EventSolver rkEventSolver) {
...
-    DES.computeDerivatives(t + h, kHelp, kVals[3]);
-    Mathematics.svMult(h, kVals[3], kVals[3]);
-    // combining all k's
-    Mathematics.svvAddScaled(2, kVals[2], kVals[3], kVals[3]);
-    Mathematics.svvAddScaled(2, kVals[1], kVals[3], kVals[2]);
-    Mathematics.svvAddAndScale(1d / 6d, kVals[0], kVals[2], change);
-    return change;
+  
+  // Using Heun's method
+  public double[] computeRK2(DESystem DES, double[] yTemp, double t, double h, double[] change)
```
TlexCypher, line 168: You need to add document string.
The methods signature written in Mathmatics package is not descriptive, so it is so hard for the user of this library to understand what each method does.
For future users, please add your thought and implementation details, some references.

TlexCypher (changes requested):
Thank you for nice contribution!
I want to understand your implementation details, so could you share some references?

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
baranwalayush: I had given it a thought. Since, in switching to default the program is not acting as per the user's input but at the same time it can continue working , that's why I went with the warning level. The info level is more general and should not include something that not looking into them could lead to potential information loss.
However, I understand that info level can also be good choice for this. Should I change it to info?

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
baranwalayush: done.

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
baranwalayush: done

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
baranwalayush: done

baranwalayush:
you can check [wikipedia](https://en.wikipedia.org/wiki/Heun%27s_method) or [libretext mathematics](https://math.libretexts.org/Workbench/Numerical_Methods_with_Applications_(Kaw)/8%3A_Ordinary_Differential_Equations/8.03%3A_Runge-Kutta_2nd-Order_Method_for_Solving_Ordinary_Differential_Equations) (Eq. 8.3.1.10, 8.3.1.11a and 8.3.1.11b) for reference.

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
TlexCypher: Since there's a clear difference in intent between a user not specifying a method (e.g. null or empty string) and providing an invalid method name, I think it would be better to handle them separately.
For example:

- If the method is missing, logging at the INFO level makes sense, since falling back to the default is expected behavior.

- If the method is invalid, a WARNING is more appropriate, as it might indicate a user mistake or a bug, and could lead to unintended behavior.

What do you think?

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
```diff
@@ -41,6 +43,19 @@ public class RungeKutta_EventSolver extends AbstractDESSolver {
...
 
+  /**
+   * Logger for this class
+   */
+  private static final Logger logger = Logger.getLogger(RungeKutta_EventSolver.class.getName());
+
+  /**
+   * Enum representing supported RungeKutta methods
+   */
+  public enum validMethod{
```
TlexCypher, line 54: Suggested changes
```java
public static enum ValidMethod {
}
```

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
```diff
@@ -81,6 +132,19 @@ public RungeKutta_EventSolver(double stepSize, boolean nonnegative) {
    */
   public RungeKutta_EventSolver(RungeKutta_EventSolver rkEventSolver) {
     super(rkEventSolver);
+    this.method = rkEventSolver.method;
+  }
+
+  /**
+   * Parses the string into a validMethod enum, defaults to RK4 if invalid.
+   */
+  private validMethod getValidMethod(String method) {
```
TlexCypher, line 141: I think the method name getValidMethod could be a bit misleading. It suggests that the method just retrieves or validates a method value, but in reality, it also falls back to a default (RK4) when the input is invalid.
A name like resolveMethod, getMethodOrDefault, or parseOrDefaultMethod might better reflect the actual behavior and make the code clearer to readers.

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
```diff
@@ -90,40 +154,128 @@ public RungeKutta_EventSolver(RungeKutta_EventSolver rkEventSolver) {
...
+
+        switch(method) {
+          case RK2:
+            return computeRK2(DES, yTemp, t, h, change);
+          
+          case RK4:
+            return computeRK4(DES, yTemp, t, h, change);
+
+          default:
+            throw new IllegalArgumentException("Unexpected method: " + method);
```
TlexCypher, line 166: You need to add supported methods name list to logging message.

TlexCypher (changes requested):

TlexCypher:
And, then could you add any unit tests?

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
baranwalayush: Hey, Sorry for the late response. Actually, I've been caught up with my mid semester exams over the past week and had a busy couple of days. 
Yes, Using INFO level for missing input and WARN for invalid input is a better approach. I'll implement it in my next commit.

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
baranwalayush: Actually, the program flow will never reach to the default case as it is already handled in the `getValidMethod` (will change the name to `parseOrDefaultMethod`). I included it just for the sake of clarity, but will add the supported methods name in the message if you prefer.

baranwalayush:
Yes, I can add unit tests. But, since the ODE solvers in SBSCL works with the SBML parsing pipeline, where any SBML model is first read by `readSBML`, then passed to `SBMLinterpreter` whose output is passed as argument to `solve` method of the solver used. To bypass all this parsing pipeline, and focus on unit testing, the best approach would be to stub the `DESystem` interface to simulate its behaviour as required. (because `DESystem` object is used as input to both the `computeChange` and `computeRK2` methods).  And then based on those dummyDESystem, I can write unit tests for the `computeRK2` method. Other interfaces like `DelayedDESystem`, `EventDESystem` are also there that themselves implements `DESystem`. But, I will keep my focus on `DESystem` and `computeRk2` only for the testing. How does that sound to you?

TlexCypher:
First of all, you should not return default method in `getValidMethod(String method)`.
Java has `throwing Exception concepts`, so if system detects invalid or illegal inputs, system should be crashed and inform it users, espasially numerical calcuration library.

I put some suggested changes, so could you tell me your thoughts?

baranwalayush:
I understand what you are trying to say. I changed the logger previously based on our discussion. Will change it now to throw exception when passed unsupported method, when no method is passed just a logger with INFO level telling that RK4 is the default method. And what do you think about the approach of writing the unit tests that I shared earlier.

TlexCypher:
Hi, @baranwalayush. Sorry for late response.
First of all, in my opinion, we don't need to make `DESystem` stub.
You need to clarify what you should test.
In this case, you only need to validate whether `computeRK2()` calculation result is correct or not.
As you said, when we want to validate this, we need to handle SBML file I/O, but this can be easily to stub, like using dummyModel.(Off cource, dummyModel shoule be formatted correctly as SBML.)
And after that you need to pass dummyModel to SBMLInterpreter, then you can check the correctness of your solver.

baranwalayush:
I believe, the existing solvers have undergone thorough integration testing and are in use from some time now. As this PR solely implements the `computeRK2` method (called by the `computeChange` method which deals with all the other operations, implemented in the `abstractDESSolver` class) without modifying the SBML parsing pipeline, concentrating on unit testing appears to be the most effective strategy (as you said, we just want to validate the `computeRK2`). I will include some failed test cases. What do you think?

TlexCypher:
@baranwalayush Thank you for response !  
I fully agree with you, I look forward it!

src/main/java/org/simulator/math/odes/RungeKutta_EventSolver.java
baranwalayush: done.

baranwalayush:
@TlexCypher,
I have incorporated all your suggestions. Could you please review it? If everything looks good, I would appreciate if you could merge it. Otherwise, let me know if there is anything else you want me to work on.

TlexCypher (approved):
LGTM

TlexCypher:
@baranwalayush 
Sorry for late response.
I give you LGTM, so merge it.
Thank you for great contribution.

PR #85, opened 2025-07-07
Update LSODA

baranwalayush:
### Description

This PR brings major updates and structural improvements to the LSODA implementation in SBSCL. Key changes include:

### Core Refactoring
- Refactored the `lsoda` driver function fixing issues and improving readability.
- Added and reorganized internal components: `allocMemory`, `checkOpt`, `lsodaPrepare`, `prepare` methods.
- A new `LSODACommon` constructor is used to manage the memory allocation within `allocMemory`.
- `lsodaPrepare` will be used only with `LSODAIntegrator` object, when an ODE system is directly passed as input.
- `Prepare` has been declared in `AbstractDESolver` and implemented in `LSODAIntegrator`, and will be used when working with SBML files.

### Exception Handling
- Added custom exceptions:  
  - `TooMuchAccuracyException` , `IllegalInputException`, `ErrorWeightException`, `IllegalTException`, `IntdyException`, `MaxStepExceededException`, `TestFailException`, `MembersNotInitializedException`, as needed.
- Updated `lsoda` to throw these exceptions at appropriate error points to improve robustness and debuggability.

### Testing
- Added `LSODASystemTest` covering multiple test scenarios:
  - Linear ODE system
  - Exponential ODE system
  - Stiff ODE system
  - SBML file as input

### Documentation
- Added detailed method-level docstrings across the `LSODAIntegrator` class.

### SBML Integration
- Implemented the `computeChange()` method to support SBML-based simulations.
- As discussed previously, designed dual input mechanisms:
  - Directly using an ODE system (via `DESystem`)
  - Parsing and using an SBML file  
 - This structure closely mirrors the original `libLSODA` design while aligning with SBSCL's architectural conventions.

---

With this PR, the core implementation of the LSODA algorithm is nearly complete. Next, I will focus on validating it against the SBML Test Suite.

### DIFF
--- a/src/main/java/org/simulator/math/odes/LSODA/LSODAIntegrator.java
+++ b/src/main/java/org/simulator/math/odes/LSODA/LSODAIntegrator.java
@@ -615,6 +615,7 @@ public class LSODAIntegrator extends AdaptiveStepsizeIntegrator {
                 }
                 common.setAcor(acor);
             }
+            // System.out.println("del = " + del[0] + " pnorm = " + pnorm);
             if (del[0] <= 100d * pnorm * common.ETA) {
                 break;
             }
@@ -1830,7 +1831,7 @@ public class LSODAIntegrator extends AdaptiveStepsizeIntegrator {
             }
 
             kflag = stepper.stoda(ctx, yOffset, jstart);
-            // System.out.println("yh -> " + common.getYh()[2][1] + ", kflag = " + kflag );
+            // System.out.println("yh -> " + common.getYh()[2][1] + ", kflag = " + kflag);
             if (kflag == 0) {
 
                 /*
--- a/src/main/java/org/simulator/math/odes/LSODA/LSODAIntegrator.java
+++ b/src/main/java/org/simulator/math/odes/LSODA/LSODAIntegrator.java
@@ -418,7 +418,7 @@ public class LSODAIntegrator extends AdaptiveStepsizeIntegrator {
                     pc[i] = pc[i - 1] + fnqm1 * pc[i];
                 }
                 pc[1] = fnqm1 * pc[1];
-                pint = pc[i];
+                pint = pc[1];
                 xpin = pc[1] / 2d;
                 tsign = 1d;
                 for (i = 2; i <= nq; i++) {
@@ -438,7 +438,7 @@ public class LSODAIntegrator extends AdaptiveStepsizeIntegrator {
                 agamq = rqfac * xpin;
                 ragq = 1d / agamq;
 
-                newTesco[nq][2] = agamq;
+                newTesco[nq][2] = ragq;
                 
                 if (nq < 12) {
                     newTesco[nqp1][1] = ragq * rqfac / (double) nqp1;
@@ -1858,7 +1858,7 @@ public class LSODAIntegrator extends AdaptiveStepsizeIntegrator {
                     if ((common.getTn() - tout) * common.getH() < 0d) {
                         //System.out.println("in itask=1");
 
-                        //System.out.println("tn = " + common.getTn() + ", h = " + common.getH());
+                        // System.out.println("tn = " + common.getTn() + ", h = " + common.getH());
                         continue;
                     }
                     //logger.log(Level.INFO, "in itask=1");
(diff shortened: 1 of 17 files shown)


src/main/java/org/simulator/math/odes/LSODA/LSODAIntegrator.java
```diff
@@ -1657,238 +1675,253 @@ public int lsoda(LSODAContext ctx, double[] y, double[] t, double tout) throws D
...
+                    }
+                    break;
+
+                case 2:
+                    break; 
+                
+                case 3:
+                    tp = common.getTn() - common.getHu() * (1d + 100d * common.ETA);
+                    if ((tp - tout) * common.getH() > 0d) {
+                        hardFailure(ctx, opt.toString(), String.format("[lsoda] itask = %d and tout behind tcur - hu", itask));
```
draeger: Can we please use `MessageFormat` rather than `String.format`? Rationale: If we ever decide to localize the application, that shift will be easier because in `MessageFormat` we can define placeholders like `{3,number,integer}` saying: the third variable goes to this position within the `String`. In contrast, `String.format` requires the order of variables to match the order of placeholders in the text, making it sometimes impossible to translate texts if we need to change the order in which variables have to be inserted to match the conventions of a particular language.

src/test/java/LSODASystemTest.java
```diff
@@ -0,0 +1,941 @@
...
+        LSODAIntegrator solver = new LSODAIntegrator();
+
+        solver.lsodaPrepare(ctx, opt);
+        solver.lsoda(ctx, y, t, tout);
+
+        double[] result = solver.getResult();
+
+        double[] res = {3.6787944436997860e-01, 6.2500152626828648e-02, 2.4691357993737381e-02, 1.1718750006733666e-02, 6.3999999971913754e-03, 
+                        3.8580246808020361e-03, 2.4989587671788559e-03, 1.7089843749999937e-03, 1.2193263222069807e-03, 9.0000000000000008e-04, 
+                        6.8301345536507063e-04, 5.3047839506169931e-4}; // result taken from liblsoda
```
draeger, line 567: Should we read these results from an external file or keep them hard-coded here?

draeger (changes requested):
I left a few comments, some of which apply to more positions than what I directly indicated, e.g., remarks about String manipulation.

src/test/java/LSODASystemTest.java
```diff
@@ -0,0 +1,941 @@
...
+        LSODAIntegrator solver = new LSODAIntegrator();
+        solver.lsodaPrepare(ctx, opt);
+
+        while(tout<=5){
+            result = Math.exp(-3 * tout);
+            solver.lsoda(ctx, y, t, tout);
+            double[] res = solver.getResult();
+
+            // System.out.println("output at " + tout + " = " + res[0] + ", expected = " + result);
+            assertTrue(Math.abs(result-res[0])<1e-6);
```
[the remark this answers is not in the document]
baranwalayush: Done.

src/test/java/LSODASystemTest.java
baranwalayush: This has been discussed in our meeting and we decided to keep it here only, as there are only a few such cases.

src/test/java/LSODAIntegratorTest.java
```diff
@@ -3243,92 +3243,92 @@ public void setDelaysIncluded(boolean delaysIncluded) {
...
+        ctx.setNeq(system.getDimension());
+        ctx.setState(1);
+        LSODAIntegrator solver = new LSODAIntegrator();
+        solver.lsodaPrepare(ctx, opt);
+
+        for(int i=1; i<=12; i++) {
+
+            solver.lsoda(ctx, y, t, tout);
+            double[] result = solver.getY();
+            System.out.println("t = " + tout + " -> " + result[1] + ",  " + result[2] + ",  " + result[3]);
```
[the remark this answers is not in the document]
baranwalayush: Already taken care of. It was in an outdated version.

src/main/java/org/simulator/math/odes/LSODA/LSODAIntegrator.java
baranwalayush: Actually I didn't knew that something like `MessageFormat` exists until now. This is definitely better than `String.format`. But, all these calls were in an outdated version and the current one doesn't have more than a couple of `String.format` calls.

TlexCypher (commented):
@baranwalayush Thank you for great contribution!
Basically, give you LGTM.
I put some nits comments.
Btw, as the next step, could you rename each variables likeillinm nhnil, nslast...?
Because that brings unreadability to our code.

funasoul:
LGTM

Issue #45, opened 2020-06-16
Misinterpretation of function variables with parameters


hemilpanchiwala:
This issue occurs when the bvars defined in the FunctionDefinition of the model are same as the defined ids for the parameters. The simulator misinterprets the function variables as the parameters and calculates the results by considering the values of the parameters.

This issue is encountered while running the SBML L3V2 model of the [1492](https://github.com/sbmlteam/sbml-test-suite/blob/develop/cases/semantic/01492/01492-sbml-l3v2.xml) test case from the SBML Test Suite. Similar is the issue for the test case [1493](https://github.com/sbmlteam/sbml-test-suite/blob/develop/cases/semantic/01493/01493-sbml-l3v2.xml).

In the test case 1492, function with id `my_piecewise` is defined as `piecewise(x, y, z)` where x, y, and z but simulator misinterprets them with the parameters defined with the same id.

Below is the screenshot of the result from the simulator for test case 1492 where z simply takes the value of parameter x (instead of some defined formula) as `piecewise(x, y, z) => x if y > 0 else z`.

![Screenshot from 2020-06-17 00-13-15](https://user-images.githubusercontent.com/42446679/84814672-702a6700-b02f-11ea-8183-535a157bc1dd.png)

dyrpsf:
Hi @hemilpanchiwala, @funasoul, @draeger, 
I’d like to work on this issue if it’s still open. I’ve gone through the problem description and the SBML test cases (1492/1493), and I believe I can help investigate and fix the variable–parameter name resolution issue in the function definition handling.

If it’s okay, please assign this issue to me. Thanks!

dyrpsf:
Hi @hemilpanchiwala, @funasoul, @draeger,
I’ve opened a pull request to fix this issue: https://github.com/draeger-lab/SBSCL/pull/96

The PR updates `ASTNodeInterpreter` so that when evaluating a `FunctionDefinition`, the function arguments (bvars) are stored in `funcArgs` before the function body is evaluated. This ensures that identifiers inside the function body correctly evaluate to the argument values, even if there are global parameters with the same ids. The previous `funcArgs` mapping is restored afterwards so nested function calls remain safe.

draeger:
Excellent, thanks for the effort, @dyrpsf!

draeger:
[...]

PR #96, opened 2026-01-04
Fix misinterpretation of function arguments with parameters (#45)

dyrpsf:
This PR fixes issue #45: when a FunctionDefinition has bvars with the same ids as
global parameters, SBSCL used the parameter values instead of the argument values
inside the function body.

The fix:
- Populate ASTNodeInterpreter.funcArgs with {bvarName -> argumentValue} before
  evaluating the function body in functionDouble(...) and functionBoolean(...).
- Restore the previous funcArgs map after evaluation, so nested function calls work.

With this change, identifiers inside a function body correctly refer to the function
arguments, even if parameters with the same ids exist in the model.

### DIFF
--- a/src/main/java/org/simulator/sbml/astnode/ASTNodeInterpreter.java
+++ b/src/main/java/org/simulator/sbml/astnode/ASTNodeInterpreter.java
@@ -153,6 +153,33 @@ public class ASTNodeInterpreter {
   }
 
   /**
+   * Evaluates function argument expressions, stores their values, and updates the
+   * {@code funcArgs} map so that identifiers inside a FunctionDefinition body resolve to
+   * the corresponding argument values. Returns the previous {@code funcArgs} map so it can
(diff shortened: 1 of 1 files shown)


draeger (approved):
This looks good: The only comment I have is that we now have two parts of the code with nearly identical and comparably complex code. If further improvements or corrections are necessary, adjustments will need to be made to the positions. If it were possible to unify both methods by performing the pre- and postprocessing steps in sub-methods, maintenance could be simplified, and duplication errors could be reduced.

dyrpsf:
Hi @draeger,
I’ve pushed a follow-up commit that factors out the common argument-handling logic into a private
`pushFunctionArguments(...)` helper. Both `functionDouble(...)` and `functionBoolean(...)` now use
this helper, so their bodies only differ in how they evaluate the function body, and the duplicated code is removed.

Issue #91, opened 2025-08-21
Raise Java baseline & ensure FBA runs cleanly on Java 22 (deprecate Java 8)


xts-Michi:
# Issue: Raise Java baseline to **17** & ensure FBA runs cleanly on **JDK 22+**

## Summary

SBSCL currently targets **Java 8**. This issue proposes to **raise the minimum required Java version to 17 (LTS)** and to **verify FBA runs cleanly on JDK 22+**. This removes Java 8 support and the related maintenance burden.

## Motivation

* Modern JDKs (17/22+) provide security, performance, and tooling improvements.
* Avoid brittle JAXP conflicts on newer JDKs (e.g., `AbstractMethodError` in Log4j2 XML config when legacy XML parsers shadow the JDK’s).
* Simplify build, CI, and contributor experience.

## Proposed Change

* **Set baseline to Java 17**: update `maven-compiler-plugin` to `<release>17</release>`.
* **CI matrix**: test on **JDK 17 (LTS)** and **JDK 22+**.
* **Classpath hardening**: remove/avoid legacy XML parser stack (`xercesImpl`, `xml-apis`, `xalan`) so the JDK’s JAXP is used.
* **Docs**: update `README.md` / `INSTALL.md` to state **Java 17+** is required.

## Observed Behavior (current)

* On JDK 22, early failures can occur in Log4j2 XML config if legacy XML parsers take precedence over the JDK’s, leading to `AbstractMethodError`.
  Removing those legacy parsers allows the FBA path to initialize and solve as expected.

## Impact / Compatibility

* **Breaking change**: downstream users must run on **Java 17+**.
* Build and runtime get simpler; fewer surprises on modern JDKs.

dyrpsf:
@xts-Michi 
Thanks for the quick response!  
I’m happy to adjust anything you’d like changed or run any additional checks.  
Just let me know what you prefer and I’ll update the PR accordingly.

PR #90, opened 2025-08-21
 Feat: set `OptSolvX` as default LP backend; add SBML-FBC→LP bridge

xts-Michi:
> **Status:** Ready for review and merge - But PR #88 should be merged before

## Summary

Switch FBA to **OptSolvX** by default, introduce an initial SBML-FBC→LP bridge, fix dependency/classpath issues for modern JDKs, **remove** the legacy SCPSolver/GLPK backend and enable configurable solver selection (Commons Math / ojAlgo) via OptSolvXConfig.

## What changed since the last review

* **Removed** `NewGLPKSolver` and SCPSolver test (`SCPSolverIssue`); no SCPSolver/GLPK deps remain in code or POM.
* Added **constraint-level tests** for the bridge (`BridgeConstraintTests`) and minor perf tidy-ups in the bridge (cached SBML lists, pre-sized maps).
* Restored `getActiveObjective()` in FBA for API/test compatibility.
* Aligned dependencies to include **Commons Math 2.2** alongside Math3 for legacy ODE API.
* Backend selection / tests verifying selection
* Added/updated bridge pipeline test to verify backend selection via -Doptsolvx.solver (ojAlgo vs commons-math).

### 1) Default backend = OptSolvX

* FBA uses OptSolvX by default via `OptSolvXSolverAdapter`.
* Solver backend is resolved via OptSolvXConfig (system property -Doptsolvx.solver=..., per-model override, fallback to commons-math).
* Supported backends: commons-math, ojalgo (aliases: cm, oj, …).

### 2) SBML-FBC → LP bridge (prototype)

* `FbaToOptSolvX` maps:

  * variables = reactions (+ bounds from FBC v1/v2),
  * constraints = **S·v = 0** with **boundary species ignored**,
  * objective = active FBC objective (v1/v2).
* Small perf pass: cache SBML lists, pre-size maps.

### 3) Dependency / classpath

* **Add** `org.apache.commons:commons-math:2.2` for legacy ODE API (Math2) next to `commons-math3:3.6.1`.
* **Exclude** legacy XML stack (`xercesImpl`, `xml-apis`, `xalan`) to avoid JAXP conflicts with modern JDKs (Log4j2 config now clean).
* OptSolvX is resolved from the local Maven repository via the `local-optsolvx` profile until it is available on Maven Central.

### 4) Cleanup: remove SCPSolver/GLPK

* Delete legacy GLPK/SCPSolver backend (code + test).
* No more SCPSolver jars on the compile/test classpath.
[...]

PR #97, opened 2026-01-05
Raise Java baseline to 17 and clean XML parsers for JDK 22+

dyrpsf:
This PR raises the Java baseline to 17 and removes legacy XML parser dependencies so SBSCL
runs cleanly on modern JDKs (tested locally on JDK 24).

Changes:
- Set `jdk.version` to 17 and updated `maven-compiler-plugin` to 3.11.0 with `<release>17</release>`.
- Excluded `xercesImpl`, `xml-apis`, and `xalan` from the `jlibsedml` dependency so that the
  built-in JDK JAXP stack is used. This avoids the AbstractMethodError seen on JDK 22+ when
  legacy XML parsers shadow the JDK implementation.
- Verified `mvn test` succeeds on a Java 17+ JDK (24 in my case); FBA code under
  `org.simulator.fba` compiles and is usable with the new baseline.
- Updated README, INSTALL, and UserGuidelines to state that Java 17+ is required.
- Updated `.travis.yml` to run the CI matrix on OpenJDK 17 and a newer OpenJDK version
  instead of Java 11.
issue #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Bio-Informatics, Libraries, Simulations
 
 ### Requirements
 
-SBSCL requires **Java 17 or newer**. Development and testing are done with JDK 17 and more recent JDKs (e.g. JDK 22+).  
+SBSCL requires **Java 17 or newer**. Development and testing are done with JDK 17 and more recent JDKs (e.g., JDK 22+).  
 Make sure your `JAVA_HOME` points to a JDK ≥ 17 before building the project with Maven.
 
 ### For users


NeumannArthur:
I agree with @draeger, since it has been nearly 5 years it should be safe to upgrade to Java 17 from a user perspective. 
I approve these changes, however I would also recommend merging and closing PR #90 first in order to avoid conflicts with @xts-Michi's work on OptSolvX, as there appear to still be some branch conflicts. This avoids any unexpected behaviour.

PR #98, opened 2026-01-06
Align ODE dependencies: add Commons Math 3.x alongside 2.x (#92)

dyrpsf:
This PR implements the first part of issue #92 by ensuring both Commons Math 2.x and 3.x
are available on the classpath:

- Added `org.apache.commons:commons-math3:3.6.1` alongside the existing
  `org.apache.commons:commons-math:2.2` in `pom.xml`.
- Verified that the project builds with `mvn -DskipTests compile` and `mvn test`.
- `mvn dependency:tree "-Dincludes=org.apache.commons:commons-math,org.apache.commons:commons-math3"`
  now lists both:
    - `org.apache.commons:commons-math:jar:2.2:compile`
    - `org.apache.commons:commons-math3:jar:3.6.1:compile`
- (Optional, if you ran it) `org.simulator.optsolvx.OptSolvXDemo` runs without errors.

Issue #92, opened 2025-08-26
Align ODE dependencies (Commons Math 2.x vs. 3.x)


xts-Michi:
## Description

SBSCL’s ODE classes import `org.apache.commons.math.ode.*` (Commons Math **2.x**), and historically the project only used **commons-math:2.2**.  

With the addition of **OptSolvX**, a newer **commons-math3** (`org.apache.commons.math3.*`) became required. Since 2.x and 3.x are not interchangeable but use different package names, both must now be present.  

A short-term fix adds **commons-math:2.2** alongside **commons-math3**, which coexist safely.

## Steps to Reproduce

1. Fresh checkout  
2. Remove Commons Math 2.2 from `pom.xml`  
3. Reload Maven  
4. Run `mvn clean package`  
5. Observe compile errors if 2.x is missing

## Expected

Project compiles and runs.

## Current Status

`pom.xml` updated to include `org.apache.commons:commons-math:2.2`.  
Build and demo run successfully.

## Follow-up

Migrate ODE code to **math3** (`org.apache.commons.math3.ode.*`) and drop 2.x once done.

dyrpsf:
Hi @xts-Michi & @draeger ,
I’ve opened a PR for this: https://github.com/draeger-lab/SBSCL/pull/98.

It adds `commons-math3:3.6.1` alongside the existing `commons-math:2.2` to `pom.xml`, and the
project builds successfully. The dependency tree now shows both Commons Math 2.x and 3.x present.

Issue #71, opened 2023-01-07
Exchangable solvers


draeger:
At the moment, a default solver (GLPK) is used as the backend for flux balance analysis. It should be possible to make this exchangable. Implement support for multiple solvers.

dyrpsf:
Hi @draeger ,
I’d like to work on this issue.

My current understanding:
- The FBA implementation currently uses GLPK as the default LP backend via SCPSolver.
- SBSCL also depends on other solver packs (e.g. LPSolve, possibly CPLEX), but these are not yet
  exposed in the FBA API.

Planned approach:
- Inspect the FBA code under `org.simulator.fba` (especially `FluxBalanceAnalysis` and related
  classes) and the example usages.
- Introduce a small abstraction (e.g. an enum or strategy) to select the LP solver backend,
  keeping GLPK as the default.
- Add an API for users to request a particular solver (e.g. GLPK, LPSOLVE, CPLEX) when creating
  an FBA instance, and route the solver construction through this abstraction.
- Add/adjust tests or examples to show using at least two different solvers.

If you have preferences about how solver selection should be configured (constructor parameter,
setter, configuration object, etc.), or about which solvers you’d most like to support first,
I’d be happy to follow your guidance.

PR #99, opened 2026-01-08
Make FBA LP backend solver exchangeable (#71)

dyrpsf:
This PR implements the “exchangeable solvers” feature for Flux Balance Analysis (issue #71).

### What this PR does

- Adds an overloaded constructor

  `FluxBalanceAnalysis(SBMLDocument doc, LinearProgramSolver solver)`

  so that any SCPSolver `LinearProgramSolver` implementation can be used as the backend
  for FBA.

- Keeps the existing constructor

  `FluxBalanceAnalysis(SBMLDocument doc)`

  but makes it delegate to the new one with `new NewGLPKSolver()`, so GLPK remains the
  default backend and existing code continues to work unchanged.

- Updates `solve()` to call the injected `solver` field instead of a hard‑coded GLPK
  instance.

### Note about Commons Math

This branch previously also contained the Commons Math 3 dependency change that belongs
to PR #98 (issue #92). That change has now been **reverted** on this branch
(commit `Revert "Add Commons Math 3 dependency alongside Commons Math 2.x (#92)"`),
so the net effect of this PR is **only** the FBA solver change.

The Commons Math 3 update itself remains confined to PR #98, which is currently on hold
as requested while the OptSolvX work is completed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -464,11 +464,6 @@
     </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
-      <artifactId>commons-math3</artifactId>
-      <version>3.6.1</version>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
       <version>3.18.0</version>
     </dependency>


dyrpsf:
Quick update: I’ve now reverted the Commons Math 3 dependency commit from this branch, so
the effective changes in this PR are only the FBA solver refactoring in `FluxBalanceAnalysis`.

The Commons Math 3 update is still present only in PR #98 (for issue #92), which will stay
on hold as requested while the OptSolvX backend is being completed. This PR (#99) can now
be reviewed independently, as it no longer changes `pom.xml`.

Issue #44, opened 2020-06-14
Event triggered before the mentioned condition


hemilpanchiwala:
The issue occurs when the `time` csymbol is used in the condition to trigger an event.

While running the SBML L3V2 test case [1444](https://github.com/sbmlteam/sbml-test-suite/blob/develop/cases/semantic/01444/01444-sbml-l3v2.xml) from the SBML Test Suite, I see that the event should trigger at the time `t >= 5` but in results, the event is triggered at the time `t = 4`.

A similar issue occurs for test cases [1445](https://github.com/sbmlteam/sbml-test-suite/blob/develop/cases/semantic/01445/01445-sbml-l3v2.xml), [1446](https://github.com/sbmlteam/sbml-test-suite/blob/develop/cases/semantic/01446/01446-sbml-l3v2.xml), [1447](https://github.com/sbmlteam/sbml-test-suite/blob/develop/cases/semantic/01447/01447-sbml-l3v2.xml), and [1448](https://github.com/sbmlteam/sbml-test-suite/blob/develop/cases/semantic/01448/01448-sbml-l3v2.xml).

Here is the screenshot of the result from the SBML Test Runner (for test case 1444)
Whenever the slope of the plot changes, at that time the event is triggered. (As per the model, it should trigger at `t = 5`)

![Screenshot from 2020-06-14 23-52-26](https://user-images.githubusercontent.com/42446679/84602427-02523400-aea5-11ea-9b5b-9e267b5671c3.png)

**Note:** Left one is the predefined result from the SBML Test Suite while the right one is the SBSCL simulator's result.

matthiaskoenig:
You should figure out where the events are evaluated in the code. I.e. somewhere it is checked in every integration step if the trigger conditions are fullfilled. Could you post the information on where to find this code then I can have a look what could go wrong.

hemilpanchiwala:
@matthiaskoenig, Are you talking about this method?
Link: https://github.com/hemilpanchiwala/SBSCL/blob/testrunner/src/main/java/org/simulator/math/odes/AbstractDESSolver.java#L731

shalinshah1993:
I think `AbsrtactDESSolver` simply runs it. Where in the code does the test make a call to the solver? What params are passed? I'm guessing that the `params` passed or the way it is called has something to do with it. It should be interesting to note that the test starts at `t = 4` and not `t = 0`. May be the parameter you passed need `t+1`?

hemilpanchiwala:
[...]

PR #103, opened 2026-02-09
Respect delay in rateOf for rate-rule variables and re-enable SBML Test Suite cases

dyrpsf:
This PR fixes the handling of `rateOf` when used inside `delay(...)` expressions and addresses #46.

### Background

For variables governed by SBML rate rules, `rateOf(x)` is evaluated by
`ASTNodeInterpreter.rateOf(...)`. The AST infrastructure already passed a `delay`
argument down to this method, but the rate rule evaluation ignored it:

```java
return eqnSystem.getRateRulesRoots().get(i).getNodeObject().compileDouble(time, 0d);
```
In other words, `rateOf` always evaluated the rate rule at `delay = 0d`, even when
called inside a `delay(...)`. This matches the behavior reported for SBML Test Suite
models 1400 and 1401, where parameters `P0` (no delay) and `P1` (with delay) had the
same values over time.

Changes

1. Respect delay when evaluating rate rules in `rateOf`

In `ASTNodeInterpreter.rateOf(...)`, for variables with rate rules, the delay
argument is now passed through to the AST node:
```java
for (int i = 0; i < eqnSystem.getRateRulesRoots().size(); i++) {
  RateRuleValue rrRoot = rateRulesRoots.get(i);
  if (rrRoot.getVariable().equals(sBase.getId())) {
    return eqnSystem.getRateRulesRoots().get(i).getNodeObject().compileDouble(time, delay);
  }
}
```
The existing behavior for variables using the `changeRate` array (ODE derivatives)
is unchanged, as is the early return for times before the delay or for constants
/ local parameters.

2. Re-enable the SBML Test Suite cases related to this bug

In `SBMLTestSuiteTest`, the following SBML Test Suite cases, which were previously
skipped with the comment "failing due to delay in rateOf (see issue #46)", have
been removed from the `failedTests` list:

  - `01400`, `01401`, `01403`, `01406`, `01409`

Testing
```bash
mvn -DskipTests compile – success
mvn test – success
```
Maintainers who have the SBML Test Suite installed should now be able to run
models 1400/1401 (and the related cases above) and observe that `rateOf` inside
`delay(...)` respects the delay for rate-rule variables.

PR #104, opened 2026-02-26
Use correct time in event triggers and re-enable SBML time-based event tests

dyrpsf:
This PR fixes the handling of SBML events whose triggers depend on `csymbol time`
and addresses #44.

### Problem

In `SBMLinterpreter.getNextEventAssignments(...)`, the time used to evaluate event
triggers, priorities, delays, and assignment rules is the field `astNodeTime`.
Previously, at each call this was advanced by a fixed epsilon:

```java
currentTime = t;
Double priority, execTime = 0d;
astNodeTime += 0.01;
```

The AST for triggers (including those using `csymbol time`) was always evaluated
at this artificial `astNodeTime`, which did not match the actual simulation time.
This led to incorrect firing times for events whose triggers depend on time,
e.g. in SBML Test Suite cases 01444–01448, where events should fire at `time >= 5`
but effectively appeared to fire earlier.

Changes

1. **Use the actual current time for AST evaluations**

In `SBMLinterpreter.getNextEventAssignments(...)`:

```java
currentTime = t;
Double priority, execTime = 0d;
// Use the actual current time when evaluating triggers, priorities, delays and rules
astNodeTime = currentTime;
```

All subsequent evaluations of triggers, priorities, delays, and assignment rules
now see the correct `time` value when `csymbol time` is used in the math.

**Re-enable the affected SBML Test Suite cases**

In `SBMLTestSuiteTest`, the following SBML Test Suite semantic test cases, which
were previously skipped with the comment "failing due to event triggers before
mentioned condition (see issue #44)", have been removed from the `failedTests` list:

- 01444, 01445, 01446, 01447, 01448

These tests are now executed again (assuming the external SBML Test Suite
resources are available).

**Tests**

- `mvn -DskipTests compile` – success
- `mvn test` – success

Maintainers who have the SBML Test Suite installed should now be able to run
cases 01444–01448 and see that events which depend on `csymbol time` fire at the
correct times.

### DIFF
--- a/src/main/java/org/simulator/sbml/SBMLinterpreter.java
+++ b/src/main/java/org/simulator/sbml/SBMLinterpreter.java
@@ -161,7 +161,8 @@ public class SBMLinterpreter extends EquationSystem {
     System.arraycopy(Y, 0, this.Y, 0, Y.length);
     currentTime = t;
     Double priority, execTime = 0d;
-    astNodeTime += 0.01;
+    // Use the actual current time when evaluating triggers, priorities, delays and rules
+    astNodeTime = currentTime;
     Double[] triggerTimeValues;
     Event ev;
     int i = 0, index;
--- a/src/test/java/org/simulator/sbml/SBMLTestSuiteTest.java
+++ b/src/test/java/org/simulator/sbml/SBMLTestSuiteTest.java
@@ -112,8 +112,8 @@ public class SBMLTestSuiteTest {
         "01287", "01592",
         // failing due to delay in rateOf (see issue #46)
         "01400", "01401", "01403", "01406", "01409",
-        // failing due to event triggers before mentioned condition (see issue #44)
-        "01444", "01445", "01446", "01447", "01448",
+        // previously failing due to event triggers before mentioned condition (see issue #44)
+        // (re-enabled after fix in SBMLinterpreter.getNextEventAssignments)
         "01456",
         "01480",
         "01481", // model below sbml l3v1


dyrpsf:
Quick update on local verification for the time-based event tests:

- I downloaded **SBML Test Suite 3.3.0** (`sbmlteam/sbml-test-suite-3.3.0.zip` from Zenodo),
- unpacked it, and placed the `cases/semantic` folder under

  `src/test/resources/sbml-test-suite/cases/semantic`

  in my SBSCL checkout,
- then ran:

  ```bash
  mvn -Dtest=SBMLTestSuiteTest -DfailIfNoTests=false test
  ```
  on the `event-trigger-issue-44` branch.

With this setup, `SBMLTestSuiteTest` finishes with `BUILD SUCCESS`, and all 1809 semantic
cases pass locally, including the time-based event tests 01444–01448 that were previously
failing.

This suggests that the current event-trigger changes are consistent with SBML Test Suite
3.3.0 when using SBSCL’s in-repo test harness.

If you’re using a different SBML Test Suite version with the external SBMLTestSuite Runner,
or have any additional logs/plots for 01444–01448, I’d be happy to try and reproduce that
setup on my side as well.

dyrpsf:
> I have the same comment as in #103: It certainly makes sense to add this expected testing procedure into a little additional section of the README file to support future contributiors.
> 
> Thanks for going this extra mile, @dyrpsf!

Thanks for the approval and for the pointer to document this in the README.

I’ve created an issue (#107 ) for a README update describing the SBML Test Suite setup and will be
happy to follow up with a small PR for that.

baranwalayush:
Added further comments in #103. Please check that review.

dyrpsf:
> Added further comments in #103. Please check that review.

Thanks again for the review and the pointers here and on PR #103.

As mentioned there, I’ve verified the fixes with `SBMLTestSuiteTest` + SBML Test Suite 3.3.0, and
I’ll read through the wrapper script and related code so I better understand the external
SBMLTestRunner pipeline as well.

PR #112, opened 2026-04-24
feat: Add LLMSimulationTraceCompressor for token-efficient AI summaries

dyrpsf:
## Overview
This PR introduces the `LLMSimulationTraceCompressor` to the `org.simulator.math` package. This is a foundational utility for the upcoming `gsoc-sysbio-llm-tools` architecture, designed to specifically solve the LLM token-limit bottleneck when handling massive simulation outputs.

## Changes Made
* **Data Compression:** Built a utility to intercept `MultiTable` time-series data and algorithmically extract critical biological waypoints (initial states, peak concentrations, peak times, and final steady states).
* **Token Optimization:** Formats the extracted data into a highly dense, LLM-readable summary string, preventing context-window overflow and reducing AI hallucinations.
* **Testing:** Added `LLMSimulationTraceCompressorTest.java` utilizing Mockito to mock `MultiTable` behaviors and verify the statistical extraction logic.

## Validation
* Compiles cleanly within the `sbscl` module.
* Validated locally alongside the full SBSCL test suite.

### DIFF
--- a/src/test/java/org/simulator/math/SimulationTraceCompressorTest.java
+++ b/src/test/java/org/simulator/math/SimulationTraceCompressorTest.java
@@ -18,18 +18,18 @@ public class SimulationTraceCompressorTest {
         Mockito.when(mockTable.getColumnCount()).thenReturn(2);
         
         // Mock time points
-        Mockito.when(mockTable.getTimePoint(0)).thenReturn(0.0);
-        Mockito.when(mockTable.getTimePoint(1)).thenReturn(1.0);
-        Mockito.when(mockTable.getTimePoint(2)).thenReturn(2.0);
+        Mockito.when(mockTable.getTimePoint(0)).thenReturn(0d);
+        Mockito.when(mockTable.getTimePoint(1)).thenReturn(1d);
+        Mockito.when(mockTable.getTimePoint(2)).thenReturn(2d);
 
         // Mock column names
         Mockito.when(mockTable.getColumnName(0)).thenReturn("Time");
         Mockito.when(mockTable.getColumnName(1)).thenReturn("Species_A");
         
         // Give Species A some data: Starts at 10, peaks at 50, ends at 5
-        Mockito.when(mockTable.getValueAt(0, 1)).thenReturn(10.0);
-        Mockito.when(mockTable.getValueAt(1, 1)).thenReturn(50.0);
-        Mockito.when(mockTable.getValueAt(2, 1)).thenReturn(5.0);
+        Mockito.when(mockTable.getValueAt(0, 1)).thenReturn(10d);
+        Mockito.when(mockTable.getValueAt(1, 1)).thenReturn(50d);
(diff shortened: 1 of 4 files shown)


draeger (approved):

Schmoho:
I would likely not frame this specifically around LLMs but more generally as a summary string mechanism.

As for the test, I am uncertain whether there is enough covered with that. At least there is a test though, thank you!

dyrpsf:
> I would likely not frame this specifically around LLMs but more generally as a summary string mechanism.
> 
> As for the test, I am uncertain whether there is enough covered with that. At least there is a test though, thank you!

You make a great point about keeping the core library naming generic. I have renamed it to `SimulationTraceCompressor` so it isn't strictly tied to LLM use cases. I also added some extra assertions to the test suite to cover edge cases like empty traces!

src/test/java/org/simulator/math/SimulationTraceCompressorTest.java
```diff
@@ -0,0 +1,58 @@
...
+    public void testGenerateSummary() {
+        // Create a Mock MultiTable
+        MultiTable mockTable = Mockito.mock(MultiTable.class);
+
+        // Define the behavior of our mock table (3 time steps, 2 columns)
+        Mockito.when(mockTable.getRowCount()).thenReturn(3);
+        Mockito.when(mockTable.getColumnCount()).thenReturn(2);
+        
+        // Mock time points
+        Mockito.when(mockTable.getTimePoint(0)).thenReturn(0.0);
```
draeger: I think it is more precise to write `1d` or `1f`, etc. rather than `1.0`. The latter will certainly be interpreted as a `double` but the first two versions give us developers freedom to specify how much memory we want to spend.

draeger (changes requested):
Minor note; I like to advice using `1d`, `1f`, `1L`, etc. to indicate the type of numeric value when working with literals.

src/test/java/org/simulator/math/SimulationTraceCompressorTest.java
dyrpsf: Done

dyrpsf:
> Minor note; I like to advice using `1d`, `1f`, `1L`, etc. to indicate the type of numeric value when working with literals.

Great tip on the explicit literal typing! I have gone through `SimulationTraceCompressorTest.java` and updated all the floating-point values to explicitly use the `d` suffix (e.g., `1d` instead of `1.0`). 

Everything is updated and the tests are passing cleanly. Let me know if there's anything else you'd like me to tweak!

Schmoho (approved):
looks good

draeger (approved):
Thanks for the hard work, @dyrpsf!

dyrpsf:
Thanks for the support and for merging this, @draeger and @Schmoho . I’m happy to make any additional changes if needed.

